PR #8, opened 2011-03-28
Implementation for tostring() / fromstring()

87:
Hi Francesc, here you go. ;-)


esc:
Um, what should we do with this one?

esc:
This is about 2 and half years old and doesn't apply to current master. I'll close it, feel free to fix it against current master and resubmit if you deem it important.

Issue #71, opened 2014-10-14
Different leftovers-size for in-core out-core carray


FrancescElies:
As suggested in #63 by @esc new issue is being opened.

For an out-of-core or in core carray containing the same data, leftovers' size should be the same.

```
In [1]: import numpy, random, bcolz, tempfile, os, shutil

In [2]: rootdir = tempfile.mkdtemp(prefix='bcolz-factorize')

In [3]: shutil.rmtree(rootdir)

In [4]: N = int(1e7)

In [5]: a = numpy.fromiter((random.choice(['NY', 'IL', 'OH', 'CA']) for _ in xrange(N)), dtype='S2')

In [6]: c = bcolz.carray(a, bcolz.cparams(clevel=5, shuffle=False, cname='blosclz'), rootdir=rootdir)

In [7]: ic = bcolz.carray(a, bcolz.cparams(clevel=5, shuffle=False, cname='blosclz'))

In [8]: c.leftovers
Out[8]: 4320133120

In [9]: ic.leftovers
Out[9]: 4367646720
```

esc:
@FrancescElies is this still relevant? last issue on the 0.10.0 milestone.

FrancescElies:
Not relevant any more, thanks

esc:
cool, thanks for following up.

PR #72, opened 2014-10-14
Fix the 'leftovers' API

esc:
Developing a solution for #71

### DIFF
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -863,10 +863,22 @@ cdef class carray:
 
     """
 
-    property leftovers:
+    property leftover_ptr:
         def __get__(self):
             # Pointer to the leftovers chunk
             return self.lastchunkarr.ctypes.data
+
+    property leftover_array:
+        def __get__(self):
+            return self.lastchunkarr
+
+    property leftover_bytes:
+        def __get__(self):
+            return self.leftover
+
+    property leftover_elements:
+        def __get__(self):
+            return cython.cdiv(self.leftover, self.atomsize)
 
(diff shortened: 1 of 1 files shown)


esc:
rebased!

CarstVaartjes:
Hi!

just was looking at what's in for 0.9 and saw this which is a really good idea, but I was wondering on how it relates to the chunk iteration? (as this is the last chunk practically, if there are actually leftovers?). But that does mean availability in cython + adjust @FrancescElies his leftover usage?
And on a sidenote, I think that a sliced view for the array is the best because you should only add things to it using the official append method right?

FrancescAlted:
Sorry, I am a bit out of context here, but my view is that if iterators or views would be used, you don't need to expose the leftovers to the user.  IMO, the leftover should be an implementation detail and converted into a private variable (i.e. leading '_').  In case you actually need this, can you describe a use case?

esc:
Yes, it is meant mostly for internal consumption, but it would be nice to refactor this part of the code so that we as developers have an easier time with it. It is something that myself and others stumble upon again and again and it would be nice to have something more intuitive. Using ``_`` and making stuff private is really a good idea and I'll do that.

FrancescElies:
Hi guys, how are you feeling about this one, would you like merge button to be pressed?

esc:
Needs to be integrated with https://github.com/Blosc/bcolz/pull/132/commits and polished (discussed) a bit.

esc:
Also, completely and entirely lacks tests, shame on me.

esc:
Closing in favour of #211

PR #81, opened 2014-10-22
try to accelerate the sum by allocating the Numpy array only once.

esc:
This is an experimental branch to show how one could re-use a numpy buffer
while iterating over chunks. Mainly it serves as a simple example for an
summary statistic algorithm and should help us refine the internal API

I encounted some weird issues with ``self.chunklen`` being in terms of rows?
Not sure what that is all about?  Anyway, I had to do some weird contortions to
get the ``nelements`` in case the dimensionality was > 1. What is the intention
here. Can we add a property to the carray to get the ``nelements`` or
``nitems``? I am sure we will need this internally all over the place as we
implement more and more algorithms.

Also, it might be nice to have something like ``_into_ndarray(self, ndarray)``
for the ``chunks`` class. Right now, one has to do some weird stuff to
decompress a chunk into an existing numpy array. This has to be easier IMHO.

Lastly, I should mention that this doesn't really speed things up
significantly in my simple tests. Presumably the bottleneck lies elsewhere.

### DIFF
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -1714,8 +1714,9 @@ cdef class carray:
 
         """
         cdef chunk chunk_
-        cdef npy_intp nchunk, nchunks
+        cdef npy_intp nchunk, nchunks, nelements
         cdef object result
+        cdef ndarray buffer_
 
         if dtype is None:
             dtype = self._dtype.base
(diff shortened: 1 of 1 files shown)


esc:
Not sure what to do with this one, related to #153 somewhat.

FrancescAlted:
Provided that @esc was not seeing any significant speed-up, it is probably better to close this one.  Feel free to re-open it if you make some progress here.

PR #84, opened 2014-11-09
carray_ext.pxd

FrancescElies:
Hi Valentin,

as previously discussed per mail, here is the carray_ext.pxd pull request.
To check everything was fine I compiled, reinstalled and ran python bcolz/tests/all.py.

I tried to make as less changes as possible, but I still have some doubts.

Refactoring dtype parameter in _adapt_dtype function, was done only to be on the safe side, but
I am not completely sure if it is strictly necessary.

These cimport ndarray, dtype, npy_intp can still be found in both files (pyx and pxd)

Looking forward to any suggestions and improvements you have.

Regards,
Francesc

### DIFF
--- a/bcolz/carray_ext.pxd
+++ b/bcolz/carray_ext.pxd
@@ -1,0 +1,60 @@
+import numpy as np
+
+from definitions cimport ndarray, dtype, npy_intp
+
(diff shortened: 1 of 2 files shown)


esc:
Closing this in favour of #98

PR #109, opened 2015-01-04
Improve the release procedure

esc:
Some obvious improvements of the relases procedure.

There are two items we should discuss in addition:

The deployment of the documentation may need to be tweaked a bit. I did set it
up to have the older docs available at:

http://bcolz.blosc.org/0.7.2/

I think this might be useful, maybe not.

Also, I don't see the value in duplicating the RELEASE_NOTES on in the wiki. In
fact I used a git clone of the wiki today and copied the release notes across
at the filesystem level.  During this I discovered, that the release notes on
the wiki had diverged(!):

https://github.com/Blosc/bcolz/wiki/Release-Notes/_compare/5d3bae0d80d8769609c4a2975f2f68a40cbfec44...ed33dff09b98ccb207a9c7e2ce68ba2d875e2dbe

If this is a good  change or not, I don't know... In any case it shows that
there is danger in duplicating the release notes to the wiki. I think people
could read them fine in the Github web interface, like so:

https://github.com/Blosc/bcolz/blob/v0.7.3/RELEASE_NOTES.rst

### DIFF
--- a/ANNOUNCE.rst
+++ b/ANNOUNCE.rst
@@ -10,7 +10,8 @@ What's new
 ``bcolz`` is a renaming of the ``carray`` project.  The new goals for
 the project are to create simple, yet flexible compressed containers,
 that can live either on-disk or in-memory, and with some
-high-performance iterators (like `iter()`, `where()`) for querying them.
+high-performance iterators (like ``iter()``, ``where()``) for querying
+them.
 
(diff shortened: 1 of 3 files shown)


esc:
Anyone have strong opinions about keeping a copy of the release notes in the wiki?

esc:
since nobody voiced any strong concerns, I'll be going ahead with my proposals.

PR #114, opened 2015-01-07
privatize get_git_description

esc:
So that it doesn't show up when auto-completing with ipython.


esc:
trivial fix, merging.

PR #124, opened 2015-01-22
Add the io 'mode' to the 'repr' for 'carray's

esc:
So one can see at a glance what one can do with this array.


esc:
Great, another weird error: https://travis-ci.org/Blosc/bcolz/jobs/47975306

esc:
great, so with blosc v1.4.1 it doesn't fail, time to bring it back into default bcolz.

esc:
Rebased and removed the debugging commits, should be fine, since master now has Blosc 1.4.1

esc:
Merging to master.

Issue #119, opened 2015-01-09
Licenses not shipped in tarball on PyPi


esc:

esc:
fixed by #143 on maint-0.8.x

PR #125, opened 2015-01-24
recursive include Blosc licences

esc:
The ``python setup.py sdist`` won't dereference symlinks so simply ship the
symlink targets too.


esc:
Fixes #119

PR #133, opened 2015-02-04
Introduce a memory_chunks class

esc:
Laying the groundstones for:

a) a fast c-based implementation of memory chunks using a c struct

b) an implementation of network_chunks or S3_chunks


esc:
Internal refactoring, should probably be expanded to become typed and accept and return the ``char *`` type.

esc:
This is will break with #153. Ideally, being able to iterate over the chunks in one way or another should be part of the public api.

CarstVaartjes:
What are memory chunks exactly? (compared to chunks as we have them now?)

esc:
I am just making the fact explicit, that ``chunks`` (in-memory) is a Python list of ``chunk`` objects, currently:

```
In [6]: b.chunks
Out[6]: 
[chunk(float64)  nbytes: 262144; cbytes: 8474; ratio: 30.94
 '[  0.00000000e+00   1.00000000e+00   2.00000000e+00 ...,   3.27650000e+04\n   3.27660000e+04   3.27670000e+04]',
 chunk(float64)  nbytes: 262144; cbytes: 10368; ratio: 25.28
 '[ 32768.  32769.  32770. ...,  65533.  65534.  65535.]',
 chunk(float64)  nbytes: 262144; cbytes: 12336; ratio: 21.25
 '[ 65536.  65537.  65538. ...,  98301.  98302.  98303.]']
```

Issue #73, opened 2014-10-14
The chunks class can not be iterated over.


esc:
Trying to do so gives some unexpected results:

```
In [21]: for chunk in c.chunks:
    print chunk
   ....: 
['NY' 'IL' 'OH' ..., 'OH' 'OH' 'OH']
['NY' 'OH' 'IL' ..., 'OH' 'OH' 'CA']
['OH' 'IL' 'OH' ..., 'NY' 'CA' 'OH']
['CA' 'OH' 'IL' ..., 'CA' 'IL' 'CA']
['OH' 'CA' 'IL' ..., 'NY' 'IL' 'IL']
['IL' 'IL' 'IL' ..., 'IL' 'OH' 'CA']
['OH' 'CA' 'OH' ..., 'NY' 'IL' 'IL']
['IL' 'NY' 'IL' ..., 'OH' 'IL' 'OH']
['OH' 'OH' 'NY' ..., 'OH' 'OH' 'IL']
['CA' 'CA' 'OH' ..., 'NY' 'IL' 'NY']
['OH' 'IL' 'IL' ..., 'NY' 'OH' 'OH']
['IL' 'CA' 'CA' ..., 'IL' 'OH' 'NY']
['OH' 'OH' 'NY' ..., 'CA' 'CA' 'IL']
['IL' 'NY' 'NY' ..., 'IL' 'OH' 'NY']
['IL' 'CA' 'CA' ..., 'OH' 'NY' 'NY']
['IL' 'OH' 'CA' ..., 'OH' 'OH' 'CA']
['IL' 'OH' 'CA' ..., 'IL' 'OH' 'NY']
['IL' 'OH' 'OH' ..., 'IL' 'OH' 'IL']
['IL' 'CA' 'CA' ..., 'CA' 'NY' 'OH']
['CA' 'CA' 'CA' ..., 'OH' 'OH' 'NY']
['NY' 'CA' 'CA' ..., 'CA' 'NY' 'NY']
['OH' 'IL' 'IL' ..., 'OH' 'OH' 'OH']
['IL' 'CA' 'CA' ..., 'IL' 'IL' 'OH']
['IL' 'IL' 'IL' ..., 'CA' 'IL' 'IL']
['IL' 'NY' 'OH' ..., 'NY' 'OH' 'IL']
['CA' 'NY' 'NY' ..., 'OH' 'OH' 'CA']
['OH' 'OH' 'CA' ..., 'OH' 'CA' 'IL']
[...]

PR #139, opened 2015-02-27
Chunk iterator

FrancescElies:
Hi,

This PR tries to make possible to write the following cython code as described in  https://github.com/Blosc/bcolz/issues/73 by @esc ```for chunk_ in chunks:```

The PR contains two commits, the first one is the iterator's implementation itself and the second one a small test script

The test script which should not e part of the PR but I included it in case someone wants to run it 
```python test_iter_chunk.py``` gave the following results
```
test_v1:	1.768 sec
test_v2:	0.0008 sec
test_v3:	0.0034 sec
```
```for chunk_ in chunks:``` is around four times slower than the previous code written by Valentin.

Is this the right way to implement it? if yes does someone have any suggestion how this could be improved?

Thanks in advance for your time

### DIFF
--- a/bcolz/carray_ext.pxd
+++ b/bcolz/carray_ext.pxd
@@ -3,7 +3,9 @@ from numpy cimport ndarray, dtype, npy_intp
 cdef class chunk:
     cdef char typekind, isconstant
     cdef public int atomsize, itemsize, blocksize
-    cdef public int nbytes, cbytes, cdbytes
+    cdef public int nbytes, cbytes, cdbytes, _chunklen
+    cdef public int _iter_count
+    cdef public ndarray _iter_values
     cdef int true_count
     cdef char *data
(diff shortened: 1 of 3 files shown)


FrancescElies:
This was an implementation for the ```chunk```  I realised what was meant was the ```chunks``` class, sorry for the misunderstanding, closing this.

esc:
This may be useful, please don't delete the branch yet!

FrancescElies:
Hi no problem, cool that this could be useful.

I am preparing another PR which modifies ```fill_chunks``` making possible the syntax you proposed ```for chunk_ in ca.chunks:```, I believe is working correctly but still need to make some checks before I can submit it, hopefully this will happen soon, but I will need some extra help to check the correctness of the modification.

Thanks!

PR #141, opened 2015-02-28
remove workaround for older cython versions and pin 0.22

esc:
See also: https://github.com/cython/cython/pull/356


esc:
Merging now.

PR #151, opened 2015-02-28
Rebased and updated version of #112

esc:

PR #140, opened 2015-02-28
Disk handling

CarstVaartjes:
Hi,

I've combined @cpcloud and @esc their improvements and added some extra checks and functionality (it should replace the older pulls + we should be able to close some issues after this) + did my long over-due auto-flush thing:

- addcol/delcol #137 <- I also added something for a use-case where you add a diskbased carray to a diskbased ctable
- purge #130
- rootdir existence check keyerror #123 <- added a check for this
- autoflush #127 (for disk-based ctable creation and append)

So I hope this completely fixes all existing issues with disk-based ctables (we actually made some workarounds ourselves which we can also remove after this pull).

BR

Carst

Issue #137, opened 2015-02-25
addcol is broken for persistent ctables


alimanfoo:
Calling addcol on a persistent ctable leaves the data in an inconsistent state, notebook with minimal code to reproduce is here:

http://nbviewer.ipython.org/gist/alimanfoo/6cc03872fe8054e63860/bcolz_ctable_addcol.ipynb

CarstVaartjes:
Atm you still need to specify a rootdir (oops), for example:

    dimension_rootdir = ct.rootdir + '/' + dimension_name
    attribute_name = natural_name(attribute_id, 'a')
        ct.addcol(list(ct[attribute_name]), name=dimension_name, rootdir=dimension_rootdir, dtype=np.int64)

There's a pull request to solve this: https://github.com/Blosc/bcolz/pull/112 
But it needs to be finished (for delcol-behaviour)

alimanfoo:
Cool, thanks, hope the PR makes it into the next release, sorry I don't
have time to help out atm. Btw I haven't looked at the PR but noticed that
with the current implementation of addcol if you pass a carray then it
won't get saved, even if you specify a rootdir in the kwargs, as carrays
are left completely untouched. IMHO a carray column should get copied into
the ctable's directory, as it does when creating a new ctable.

On Wed, Feb 25, 2015 at 11:02 PM, Carst Vaartjes <notifications@github.com>
wrote:

> Atm you still need to specify a rootdir (oops), for example:
>
> dimension_rootdir = ct.rootdir + '/' + dimension_name
> attribute_name = natural_name(attribute_id, 'a')
>     ct.addcol(list(ct[attribute_name]), name=dimension_name, rootdir=dimension_rootdir, dtype=np.int64)
>
> There's a pull request to solve this: #112
> <https://github.com/Blosc/bcolz/pull/112>
> But it needs to be finished (for delcol-behaviour)
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/Blosc/bcolz/issues/137#issuecomment-76080698>.
>



-- 
Alistair Miles
Head of Epidemiological Informatics
Centre for Genomics and Global Health <http://cggh.org>
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Web: http://purl.org/net/aliman
Email: alimanfoo@gmail.com
Tel: +44 (0)1865 287721

esc:
[...]

PR #130, opened 2015-02-02
Implement ``carray.purge``

esc:
Purge was the best name I could think of, alternative suggestions welcome.

PR #152, opened 2015-02-28
Disk based (updated #140 based on #151 + more atomic commits)

CarstVaartjes:
Updated #140 for the new master, with @esc his changes in #151 and more atomic commits. This should solve all known issues with handling disk-based ctables & carrays

addcol/delcol #137 <- I also added something for a use-case where you add a diskbased carray to a diskbased ctable
purge #130
rootdir existence check keyerror #123 <- added a check for this
autoflush #127 (for disk-based ctable creation and append)

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -7,6 +7,8 @@ Changes from 0.8.1 to 0.9.0
 
 - Implement ``purge``, which removes data for on-disk carrays. (#130 @esc)
 
+- Implement addcol/delcol to properly handle on-disk ctable (#112 @cpcloud)
+
 Changes from 0.8.0 to 0.8.1
 ===========================
 
--- a/bcolz/ctable.py
+++ b/bcolz/ctable.py
@@ -518,10 +518,7 @@ class ctable(object):
         # Remove the column
         col = self.cols.pop(name)
 
-        # remove the data if we have a rootdir
-        if self.rootdir is not None and col.rootdir is not None:
-            # is the col.rootdir is not None check necessary?
-            shutil.rmtree(os.path.join(self.rootdir, name))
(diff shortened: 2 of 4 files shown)


CarstVaartjes:
Btw, i'm not sure what would be the best way to add tests for these; smart copying would be straight forward i guess (check the paths and validate the ctable) but the other ones would be rather indirect. For autoflush, close the ctable, reopen it and see if the changes were saved? Ideas / help are welcome here :)

esc:
+1 on the flushing.

esc:
Regarding the tests:

For the smart copying, yes, checking that the files have moved or not moved is probably good.

For the autoflush, you probably want to do what you suggested, but you could also use the ``mock`` package (https://pypi.python.org/pypi/mock ) to mock away the ``flush`` and ensure that it is called appropriately. This has the added benefit of being able to check that flush was called, even if it is a no-op, i.e. when calling it doesn't have any effect. I don't think mock is already a test dependency, since it isn't being used in this project (yet) but that isn't a problem.

FrancescElies:
I have a general question, shall the feature autoflush be triggered for a carray in the same way as when appending data to a ctable?

CarstVaartjes:
Thanks @FrancescElies! @esc: is it okay like this? for us it would be a significant improvement at least (removes several workarounds that we now had to have when adding cols to a disk-based ctable)

esc:
Not sure why, but Travis-CI failed to launch for this one. Perhaps because there would have been merge-conflicts against master...

Anyway, being the git-nerd I am, I did some history beautification and conflict resolution:

https://github.com/Blosc/bcolz/pull/154

Let's see what Travis-CI says.

esc:
Merged via #154

PR #159, opened 2015-03-08
install carray_ext.pxd with the package

ARF1:
This allows other packages to be cython compiled against the installed bcolz
package. Use case: building bquery

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -19,6 +19,8 @@ Changes from 0.8.1 to 0.9.0
   (``mv``) into the root directory of the ctable.
   (#140 #152 @FrancescElies @CarstVaartjes)
 
+- Distribute ``carray_ext.pxd`` as part of the package. (#159 @ARF)
+
 Changes from 0.8.0 to 0.8.1
 ===========================
 
--- a/setup.py
+++ b/setup.py
@@ -234,5 +234,5 @@ for binary data.
                     extra_compile_args=CFLAGS),
       ],
       packages=['bcolz', 'bcolz.tests'],
-
+      package_data={'bcolz': ['carray_ext.pxd']},
 )


esc:
We had been discussing this on the mailinglist and if it is a viable solution I'd be very happy to include it. Can you explain a bit more how it works? I mean, how does Cython find this? Are any additional things required?

ARF1:
@esc All that is required is the single line addition to `setup.py` that is contained in the pull request. It tells distutils to not only package the usual files, but also `carray_ext.pxd`.

About how the `carray_ext.pxd` is then used: in cython one can use e.g. `from bcolz.carray_ext cimport carray, chunks, chunk` to access directly the bcolz cython layer without passing through python.

Does this answer your questions?

esc:
So the path that Cython searches is the usual PYTHONPATH?

esc:
cc @FrancescElies @CarstVaartjes you guys OK with this one?

ARF1:
@esc Importing in cython with `cimport` behaves exactly the same way as python `import`. - At least as far as I know.

So cython will find `carray_ext.pxd` for `cimport bcolz.carray_ext` in the bcolz module directory, just like python will find `ctable.py` for `import bcolz.ctable`.

esc:
Sweet!

CarstVaartjes:
sounds really great! (learning a lot here)

FrancescElies:
very nice!

esc:
Okay, please add an entry to ``RELASE_NOTES.rst`` and I'll merge it up. You can get an idea of the format by looking at the contents of the file.

ARF1:
@esc  Great, I updated the `RELEASE_NOTES.rst` as requested.

esc:
Thanks for solving this issue, much appreciated!

PR #162, opened 2015-03-19
handle trivial boolean expressions in ctable.__getitem__() gracefully

ARF1:
Trivial boolean expressions supplied to `ctable.__getitem__()`, i.e. `True` or `False`, or even `(True & False)`, should be handled gracefully.

Example:
```
db = bcolz.ctable(...)
db['False']
```
will return an empty result and similarly:
```
db['True'] == db['(True)'] == db['(((True)))'] == db[:]
```

This PR will simplify the development of query optimizers building on bcolz that restructure complicated query expressions. If an optimizer determines a complicated query to evaluate to logical True/False (e.g. using information from a cached bquery factorization), it can pass this result to bcolz to obtain an appropriately instantiated result.

### DIFF
--- a/bcolz/ctable.py
+++ b/bcolz/ctable.py
@@ -1131,8 +1131,15 @@ class ctable(object):
                     raise IndexError(
                         "`key` %s does not represent a boolean "
                         "expression" % key)
-                return self._where(arr)
-            return self.cols[key]
+                elif arr == False:
+                    dtype = np.dtype([(name, self.cols[name].dtype) for name in self.names])
+                    return np.empty(0, dtype=dtype).view(np.ndarray)
+                elif arr == True:
+                    start = stop = step = None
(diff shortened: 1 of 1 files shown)


ARF1:
If this PR is acceptable in principle, I would amend it with similar fixes for `ctable.eval()`, `carray.__getitem__()` and `carray.eval()`.

esc:
Could you add some minimal tests, perhaps converting the examples you cite here into test-cases?

ARF1:
@esc I am working on a new approach for this PR. Should I close until it is ready?

esc:
Do whatever you feel like. Progress is slow right now and it doesn't make any difference if this issue is open or not. Just take you time.

esc:
If you have a new one, please submit it.

PR #163, opened 2015-03-20
Add safe attribute to carray

mrocklin:
This bypasses type, dtype, and stride checking in to_ndarray.
This is unsafe behavior for general users but removes significant
overhead in the "many small writes" case typical of shuffling data.

### DIFF
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -944,7 +944,7 @@ cdef class carray:
             return len(self.shape)
 
     property safe:
-        "The size of this object."
+        "Whether or not to perform type/shape checks on every operation."
         def __get__(self):
             return self._safe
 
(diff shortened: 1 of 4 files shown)


mrocklin:
If you all are ok with this idea it would be good to get other suggestions where this could be applied.  It looks like some of the logic in `append` might also be skippable.

mrocklin:
FWIW I'm using this here https://github.com/ContinuumIO/dask/pull/80

I'm happy to do whatever work necessary to get this in.  Hopefully this change is small enough to be easy to review.

mrocklin:
Any thoughts on this?  I can do without it if necessary but I'd like to know if I should or should not expect something like this to get in.

bcolz/carray_ext.pyx
```diff
@@ -939,6 +943,11 @@ cdef class carray:
         def __get__(self):
             return len(self.shape)
 
+    property safe:
+        "The size of this object."
```
FrancescAlted: The docstring needs some correction.

mrocklin: Fixed

FrancescAlted:
+1 for merging

mrocklin:
Thanks!

esc:
I updated the labels, milestone and am missing the entry to the RELEASE_NOTES.rst. Recently I had requested people to add that as part of the pull-request.

Issue #22, opened 2012-09-13
Make private methods of carray/ctable actually private


FrancescAlted:
In last 0.5 release several public methods in carray and ctable were actually meant to be private:

In [36]: c.  # a carray
c.append         c.copy           c.dtype          c.len            c.ndim           c.reshape        c.size           c.wheretrue      
c.attrs          c.cparams        c.fill_chunks    c.mkdirs         c.next           c.resize         c.sum            c.write_meta     
c.cbytes         c.create_carray  c.flush          c.mode           c.open_carray    c.rootdir        c.trim           
c.chunklen       c.dflt           c.iter           c.nbytes         c.read_meta      c.shape          c.where          

In [35]: t.   # a ctable
t.addcol         t.cbytes         t.cparams        t.dtype          t.iter           t.mode           t.ndim           t.rootdir        t.trim
t.append         t.cols           t.create_ctable  t.eval           t.len            t.names          t.open_ctable    t.shape          t.where
t.attrs          t.copy           t.delcol         t.flush          t.mkdir_rootdir  t.nbytes         t.resize         t.size

esc:
@FrancescAlted I'd like to tackle this one next. What are you current thoughts?

esc:
#209 has been merged, closing this one.

(some documentation upgrades are still needed..)

FrancescElies:
If you feel I could helpful with the documentation, please let me know

FrancescAlted:
Glad that you asked :)  Yes, documentation is very important albeit not many developers like to do it.  I suggest to produce a nice documentation addressed to developers of packages designed to use bcolz as part of the infrastructure (and I think you have quite a lot of practical skills here ;).

FrancescElies:
I'll be glad to help with that :), I opened a new PR https://github.com/Blosc/bcolz/pull/214 for the moment only with two TODOs, if anybody has any suggestions you could write them over there

esc:
I was thinking about iterating over the docstrings one more time, since we now have the public/private seperation in a better state., For the docstrings of the cython methods, special care is needed. IIRC you need the signature as the first line of the docstring, because Python can't introspect the signature of c-extension functions. Not sure though, I'll take a look later on today perhaps.

esc:
I just recall not finishing the docstring overhaul the last time, because the public/private seperation wasn't done yet.

esc:
Also overhauling the docs to use the with statement would be nice.

esc:
[...]

PR #171, opened 2015-03-31
Refactor how the documentation is built.

esc:
It turns out, that very little of the class and module documentation was
automatically extracted from the Python docstrings. Presumably this is due to
the fact that sphinx + numpy style docstrings have been a bit of a pain to work
with historically. I tried again and it seems that nowadays it actually works
fairly reasonably.

This removes a large amount of technical debt and also potential for
duplication. I didn't take a very close look, but I did see some of the
docstrings already out of sync, so fetching the ones from the source code is
usually a good idea. At least it means we won't need to maintain two copies in
future.

One issue that remains is that Cython functions may not be introspected, so the
signature of each function must be the first line of the docstring. While this
still is duplication it isn't too bad since the actual signature and the
documented signature are very, very close together.

Anyway, I wanted to put this up, to make sure we don't duplicate developer
efforts. (just in case, someone else was wanting to dig into this)


FrancescAlted:
:+1:

Enviat des del meu iPhone

El 31/03/2015, a les 21:52, Valentin Haenel <notifications@github.com> va escriure:

esc:
In light of #22 maybe I shouldn't waste time on carray now and wait for the refactoring to be resolved first.

esc:
FWIW: I tried to hookup readthedocs but bcolz can't be built there yet, because the build system and dependency handling is not entirely compatible.

In particular Cython isn't being installed, because it isn't listed anywhere in a place where dependencies are installed automatically (setup.py or requirements.txt).

esc:
here goes.

PR #172, opened 2015-04-01
Resolves #136

### LINKED ISSUE #136: ctable dtype is broken for tables with multidimensional columns
Apologies I don't have time to submit a PR with tests right now, but I believe the dtype property on the ctable class does not correctly handle the case where a table column has more than one dimension. This causes an attempt to load data from a ctable into a numpy array via `__getitem__` to fail.

To work around this I am currently monkey-patching the ctable.dtype property with the following code:

```python
def _ctable_dtype(self):
    """The data type of this object (numpy dtype)."""
    names, cols = self.names, self.cols
    l = []
    for name in names:
        col = cols[name]
        if col.ndim == 1:
            t = (name, col.dtype)
        else:
            t = (name, (col.dtype, col.shape[1:]))
        l.append(t)
    return np.dtype(l)

bcolz.ctable.dtype = property(_ctable_dtype)
```

alimanfoo:
This PR resolves #136 by modifying the ctable dtype to support multidimensional columns. Included are test cases for attempting to retrieve a total slice ([:]) from a ctable, which previously failed if the ctable had a column with >1 ndim.

### DIFF
--- a/bcolz/ctable.py
+++ b/bcolz/ctable.py
@@ -146,7 +146,15 @@ class ctable(object):
     def dtype(self):
         "The data type of this object (numpy dtype)."
         names, cols = self.names, self.cols
-        l = [(name, cols[name].dtype) for name in names]
+        l = []
+        for name in names:
+            col = cols[name]
+            if col.ndim == 1:
+                t = (name, col.dtype)
+            else:
+                # column is multidimensional
+                t = (name, (col.dtype, col.shape[1:]))
+            l.append(t)
         return np.dtype(l)
 
     @property
--- a/bcolz/ctable.py
+++ b/bcolz/ctable.py
@@ -149,11 +149,9 @@ class ctable(object):
         l = []
         for name in names:
             col = cols[name]
-            if col.ndim == 1:
-                t = (name, col.dtype)
-            else:
-                # column is multidimensional
-                t = (name, (col.dtype, col.shape[1:]))
+            # Need to account for multidimensional columns
+            t = (name, col.dtype) if col.ndim == 1 else \
+                (name, (col.dtype, col.shape[1:]))
             l.append(t)
         return np.dtype(l)
 
--- a/bcolz/tests/test_ctable.py
+++ b/bcolz/tests/test_ctable.py
@@ -505,6 +505,34 @@ class getitemTest(MayBeDiskTest):
         assert_array_equal(t[colnames][:], ra2,
                            "ctable values are not correct")
 
+    def test05(self):
+        """Testing __getitem__ with total slice"""
+        N = 10
+        ra = np.fromiter(((i, i * 2.) for i in xrange(N)), dtype='i4,f8')
+        t = bcolz.ctable(ra, rootdir=self.rootdir)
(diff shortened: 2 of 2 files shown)


bcolz/ctable.py
```diff
@@ -146,7 +146,15 @@ def cparams(self):
...
-        l = [(name, cols[name].dtype) for name in names]
+        l = []
+        for name in names:
+            col = cols[name]
+            if col.ndim == 1:
+                t = (name, col.dtype)
+            else:
+                # column is multidimensional
+                t = (name, (col.dtype, col.shape[1:]))
+            l.append(t)
```
esc: Suggest:

```[(name, col.dtype) if col.ndim == 1
                                 else (name, (col.dtype, col.shape[1:]))
     for name, col in ((name, cols[name]) for name in names)]```

or something like that... maybe?

Comment (esc) [inline bcolz/ctable.py]: But perhaps that is contrived.

**Comment by esc** [comment]
In any case +1, I guess.

**Comment by alimanfoo** [comment]
Happy to go with whichever you prefer.

On Wednesday, 1 April 2015, Valentin Haenel <notifications@github.com>
wrote:

> In any case +1, I guess.
>
> --
> <https://github.com/Blosc/bcolz/pull/172#issuecomment-88584146>.
>


-- 
Alistair Miles
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@gmail.com
Tel: +44 (0)1865 287721

**Comment by esc** [comment]
@FrancescAlted what do you say?

**Comment by esc** [comment]
After looking at it again, the thing that bums me a little is that it's not really a unit-test per se, since it doesn't test the ``dtype`` directly. I would envision a test that creates a bunch of different ctables each with a desired ``dtype`` and checks that it is set correctly.

**Comment by alimanfoo** [comment]
FWIW I don't think it's necessary to test the dtype directly, this seems
well covered by retrieving a total slice ('[:]') from the ctable and then
calling assert_array_equal.

Arguably the tests I've added could be moved under the createTest class
instead.

Happy to implement what you think is best, please provide specific
instructions.

On Wednesday, April 1, 2015, Valentin Haenel <notifications@github.com>
wrote:

> After looking at it again, the thing that bums me a little is that it's
> not really a unit-test per se, since it doesn't test the dtype directly.
> I would envision a test that creates a bunch of different ctables each with
> a desired dtype and checks that it is set correctly.
>
> --
> <https://github.com/Blosc/bcolz/pull/172#issuecomment-88604836>.
>


-- 
Alistair Miles
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@gmail.com
Tel: +44 (0)1865 287721

File: bcolz/ctable.py (hunk 1 above)
Comment (esc) [inline bcolz/ctable.py]: After pondering on this some more, I think the best would be:

```
t = (name, col.dtype) if col.ndim == 1 else (col.dtype, col.shape[1:])
```

esc:
Just the last fix and I'll get it merged up.

esc:
Thanks!

PR #178, opened 2015-04-23
allow adding strings > len 1 to ctable

brentp:


bcolz/ctable.py
```diff
@@ -371,7 +371,10 @@ def append(self, cols):
             if sclist and not hasattr(column, '__len__'):
```
esc: This code is a bit convoluted, can you add ``or isintance(column, _strtypes)`` to this line, rather than the many lines below, or alternatively, use:

```
clen = len(column) if not isinstance(column, _strtypes) else 1
```
or:

```
clen = 1 if isinstance(column, _strtypes) else len(column)
```

bcolz/tests/test_ctable.py
```diff
@@ -2081,6 +2081,17 @@ def test_08(self):
 
         t.flush.assert_called_with()
 
+    def test_strings(self):
+        """Testing that we can add fixed length strings to a ctable"""
+        dtype = np.dtype([("a", "|S5"),
+                   ("b", np.uint8),
```
esc: please align the record entries here.

esc:
@FrancescAlted any comments on this one?

esc:
Also, the test isn't Python-3 compliant.

esc:
Looks good, merging now.

Issue #66, opened 2014-10-10
Consider adding a Categorical datatype to bcolz


esc:
To be inspired by:

https://pandas-docs.github.io/pandas-docs-travis/categorical.html

ARF1:
I would be quite interested in this as well.

Rationale: pandas & bcolz string columns is currently extremely slow, as each entry in the string column is converted to a python string object.

It does not appear that a fixed length pandas dtype is likely to be available any time soon since the work required seems to be non-negligible: pydata/pandas#5261

@mrocklin is working around this deficit in his dask.dataframe project by using the pandas 'category' dtype.

The downside to this solution is that when reading a bcolz string column, the column has to be factorized for every read. It would be nice if the factorization could be stored with the data.

The [visualfabriq/bquery](http://github.com/visualfabriq/bquery) project has the "machinery" required for a categorical dtype in place. As mentioned the api for categorical access is still missing.

How should the access functions behave?

My gut says:
 - keep the factorization (and un-factorization) machinery external (use pandas, bquery, or similar)
 - getitem:
  - for numpy output type: return the factor & provide a new function to access the category data
  - for a (new) pandas output type: return a dataframe / series with pandas dtype 'category'
 - setitem: similarly

Does this sound like a sensible idea?

Is anybody else interested in this issue as well?

mrocklin:
There is a trade off here.  I think that categoricals are fantastic.  However I also appreciate that bcolz has a simple data model that exactly mirrors NumPy.

I like the `bquery` approach of building projects on top of `bcolz` because it allows bcolz itself to remain very simple.

esc:
I think the categorical type can be implemented as an additional type in an external library, given the Cython-API it would be possible.

Some notes on the implementation:

The performance of a categorical type depends very much on the "load-factor" i.e. the ratio of unique types to total number of elements. This influences how to store it in-memory and out-of-core.
I am note yet sure how to store the categories and if we should have a two way lookup. 

Depending on the number of categories, it may be wise to deactivate the shuffle filter. For example if you "only" have < 256 categories, it may or may not make any difference. Anything beyond that depends on both the entropy and the order of the data. For example, if you have 257 categories but the element that maps to integer 256 appears only once in your >> 257 elements dataset, using the shuffle filter should give some nice compression ratios. If on the other hand your coresponding values are sampled uniformly from the closed interval 0-65536 (short/int16) then the shuffle filter is unlikely to give you much advantage.

[...]

PR #184, opened 2015-05-03
pandas out_flavor for ctable

### LINKED ISSUE #176: Pandas out_flavor for better ctable performance
In this issue I want to make the case for the extension of the effect of `out_flavor` to `__getitem__()` (and related functions) and the introduction of a `pandas` outflavor. While I appreciate the rationale for limiting bcolz to the numpy data model I believe the possible performance improvements with pandas merit consideration.

I would be very interested to know if this has any chance for inclusion in bcolz, since the effort would be non-trivial for a clean implementation.

### Executive Summary ###
 - bcolz is a columnar (albeit chunked) data container
 - thus after decompression of each chunk, the data for each column is (chunk-)contiguous
 - currently however, this column-contiguous data is forced into a row-major memory structure (numpy structured array)
 - this has two consequences:
   1. the contiguous column-data has to be copied in row-sized chunks which is much slower than en-bloc copying of contiguous data,
   2. the data has to be copied (rather than being decompressed directly into the "proper" place)
 - Moving to a column-major memory layout (e.g. pandas DataFrame) would provide:
   1. much faster assignment of data to colums (which is a very common operation with `ctable`)
   2. opens the possibility for avoiding repeated copy operations in the first place
   3. ctable would be able to really leverage multi-core machines. Currently the numpy structured array is a bottleneck for `__getitem__()` and the like: #174
   4.  improve down-stream performance of data analysis as much data analysis is carried out along column rather than along rows
 -  due to the similar data access model of numpy structured arrays and pandas dataframes most code could probably remain unchanged after the output data object is initialised either as `numpy` or `pandas`
 
### Evidence of column assignment bottleneck ###
Note: for the sake of simplicity, when I use the term "column" in relation to numpy structured arrays, I refer to the fields of the (single column) structured array that bcolz uses to store its output column.

Numpy structured arrays (as used by bcolz) are inherently row-major. It is impossible to change this as far as I can see. This means that column assignment is fairly slow:
```
In [1]: import numpy as np

In [2]: M,N=int(1e7),10

# make a row-major numpy array

In [4]: A1=np.zeros((M,N),'f')

In [9]: dt=np.dtype(','.join(['f' for _ in range(N)]))

# make a structured numpy array

In [10]: A2=np.zeros((M,),dtype=dt)

In [11]: X=np.arange(M+0.0)

### simulate ctable __getitem()__ operation:
# assignment to a row-major numpy array column for column

In [13]: %timeit for n in range(N):A1[:,n]=X
1 loops, best of 3: 2.36 s per loop

# assignment of structured array field for field

In [15]: %timeit for n in dt.names: A2[n]=X
1 loops, best of 3: 2.36 s per loop
```

Significant speedup (factor 6.5!) can be achieved by moving to a column-major memory layout for the numpy array:
```
In [1]: import numpy as np

In [2]: M,N=int(1e7),10

# make a colum-major numpy array
In [3]: A1=np.zeros((M,N),'f', 'F')

In [4]: dt=np.dtype(','.join(['f' for _ in range(N)]))

# make a structured numpy array

In [5]: A2=np.zeros((M,),dtype=dt)

In [6]: X=np.arange(M+0.0)

### simulate ctable __getitem()__ operation:
# assignment to a column-major numpy array column for column

In [8]: %timeit for n in range(N):A1[:,n]=X
1 loops, best of 3: 374 ms per loop

# assignment of structured array field for field

In [9]: %timeit for n in dt.names: A2[n]=X
1 loops, best of 3: 2.43 s per loop
```

### Why Pandas DataFrames would help ###
While moving to column-major numpy arrays would be the ideal solution, this is obviously not an option: they require the entire array to have a homogeneous dtype.

Pandas DataFrame however supports columns of different dtypes and by design stores data in column-major order. (As I remember because Wes McKinney said that most of his data analysis happened along columns, though I cannot find the reference. That said I think the following article in which he explains his reasons for not choosing numpy structured arrays is interesting: [Wes McKinney: A Roadmap for Rich Scientific Data Structures in Python](http://wesmckinney.com/blog/a-roadmap-for-rich-scientific-data-structures-in-python/))

In addition, the choice of column-major ordering permits size-mutability: columns can be added without copying the data. This fits well with the ctable column-store philosopy. With numpy structured arrays, the entire array has to be copied (almost entry by entry as far as I can see) to add a new column.

Due to these advantages, column-major ordering is used my many well-known dedicated number crunching environments, among others: Fortran, MATLAB and R

Note: Making a pandas DataFrame for an already existing numpy structured array returned by ctable is not an option either: again, effectively element by element copying of the data is required.

### Evidence of performance improvements with Pandas ###

Instantiation of Pandas DataFrames is admittedly an issue with smaller databases. Leaving this issue aside for the moment, one can see that assignment to an (instantiated DataFrame) is much faster:

```
In [1]: import numpy as np

In [2]: import pandas as pd

In [3]: M,N=int(1e7),10

In [4]: A1=np.zeros((M,N),'f', 'F')

In [5]: dt=np.dtype(','.join(['f' for _ in range(N)]))

In [6]: A2=np.zeros((M,),dtype=dt)

In [7]: X=np.arange(M+0.0)

# make a pandas DataFrame

In [8]: df_A1 = pd.DataFrame(index=xrange(M), columns=range(N), dtype='f')

### simulate ctable __getitem()__ operation:
# assignment to a column-major numpy array column for column

In [9]: %timeit for n in range(N): A1[:,n]=X
1 loops, best of 3: 369 ms per loop

# assignment to a structured numpy array field for field

In [10]: %timeit for n in dt.names: A2[n]=X
1 loops, best of 3: 2.36 s per loop

# (optimized) assignment to a pandas DataFrame column for column

In [8]: %timeit for n in range(N): df_A1._data.blocks[0].values[n,:]=X
1 loops, best of 3: 364 ms per loop
```

Instantiation of the DataFrame will probably be an issue. My gut reaction is that it should be possible to instantiate (and cache) an empty DataFrame with the correct structure and then for each `__getitem__()` call shallow-copy this template and assign new data arrays. This should help with the instantiation overhead, since the column makeup of a ctable instance usually does not change too often during a program.

### Down the road ###
In the long run it might be worth getting rid of the memory copies altogether for DataFrame out_flavor and decopressing the chunks directly into the arrays backing the DataFrame. This would likely lead to further performance improvements (though smaller as the remaining memory copies would already be efficient en-bloc copies).

ARF1:
Closes #176.
Simplifies implementation of #66.

Summary:
* introduction of an abstraction layer for the "output array"
* implementation of an numpy specialisation of the abstraction layer
* implementation of a pandas specialisation of the abstraction layer

This is a quick hack to demonstrate the possible performance gains by using a output flavor with column major ordering, here: the pandas dataframe.

The architecture would need to be improved upon since this implementation suffers a x3-4 performance penalty for db[1] -type queries due to increased python overhead. For queries returning a larger number of rows this penalty disappears.

Timing results in #176.


FrancescAlted:
Would you mind to add some benchmarks in the 'bench/' directory showing the advantage of this approach?  My idea is to setup a speed regression check based on different benchmarks there.
Thanks!

ARF1:
@FrancescAlted 
> Would you mind to add some benchmarks in the 'bench/' directory showing the advantage of this approach? 

I would be happy to. I just need to clarify what you are looking for:

This PR (pandas out_flavor) was only intended as a proof-of-concept, it was not really intended for inclusion in the code-base. The architecture of the more general #187 (abstraction layer) is more performant (and easier to read).

Would you like me to provide a sample implementation of a pandas "out_flavor" for the new #187 (abstraction layer) instead and a benchmark for that? I.e. with a benchmark in analogy to `bench\getitem.py`.

Or would you like a "rawer" benchmark, avoiding `__getitem__()` (and its overhead) showing only the best possible performance for filling a pandas dataframe? Sort of like `bench\pandas-todataframe.py` does?

ARF1:
@FrancescAlted On reflection, I probably was not as clear as I could have been: when you speak of "this approach", do you mean

* the column-major (vs. row-major) result array in isolation or 
* the abstraction layer (in whatever version) plus the pandas out flavor implementation (vs. the current non-abstracted out flavor)?

esc:
What do you want us to do with the pull-request?

PR #187, opened 2015-05-05
introduction of an abstraction layer for the "results array"

### LINKED ISSUE #66: Consider adding a Categorical datatype to bcolz

ARF1:
Implicitly closes #66 and #176 by allowing users to bring their own implementations.

This introduces an abstraction layer for the generation of the "results array" in `ctable.__getitem__()` and data access.

It minimizes impact on core `bcolz` code and minimizes overhead using lookup optimizations:

Without abstraction layer:
```
In [3]: a = bcolz.open(rootdir='testdata.bcolz')

In [4]: %timeit a[1]
10000 loops, best of 3: 39.6 µs per loop
```

With abstraction layer:
```
In [3]: a = bcolz.open(rootdir='testdata.bcolz')

In [4]: %timeit a[1]
10000 loops, best of 3: 43 µs per loop
```


bcolz/defaults.py
```diff
@@ -71,6 +72,27 @@ def eval_out_flavor(self, value):
...
+    @ctable_out_implementation.setter
+    def ctable_out_implementation(self, value):
+        if value is None:
+            value = OutputStructure_numpy
+        try:
+            bcolz.ctable._outstruc_allocate = value.allocate
+            bcolz.ctable._outstruc_fromindices = value.fromindices
+            bcolz.ctable._outstruc_fromboolarr = value.fromboolarr
+            assert hasattr(value, '__setitem__')
+        except AttributeError, AssertionError:
```
FrancescAlted: Parenthesis are needed for Python 3

FrancescAlted:
I need to think a bit more on this, but at least initially this seems like a good idea to me.

esc:
Tests?

FrancescAlted:
I think that this would definitely be a nice addition.  We still need tests and update the documentation so that user "may bring their own implementation".  @AR1 could you please do that?

PR #189, opened 2015-05-06
cache ctable.dtype

ARF1:
Among others, this allows improving the performance of custom implementations of the ctable out_flavor abstraction layer #187.


FrancescAlted:
Could you please send an estimate on the kind of speedup this buys for your use case?  Thanks!

PR #190, opened 2015-05-17
Minor: docs still refer to carray name

esc:
"The carray package comes with a handy object that arranges data by column (and not by row, as in NumPy’s structured arrays). This allows for much better performance for walking tabular data by column and also for adding and deleting columns."

http://bcolz.blosc.org/tutorial.html#tutorial-on-ctable-objects


esc:
Transformed issue into pull-request.

esc:
thx

PR #196, opened 2015-05-20
simple 'to numpy array' wrapper

esc:
I keep looking for this in the docs, until I realize one needs to do a '[:]'

### DIFF
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -526,6 +526,9 @@ cdef class chunk:
             return array[::step]
         return array
 
+    def to_numpy_array(self):
+        return self[:]
+
     @property
     def pointer(self):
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Hmm, I think this wrapper is unnecessary.  Unless this `to_numpy_array()` is kind of a standard, I'd rather add a section in the documentation on how to get NumPy arrays out of carray/ctable objects.

esc:
Well, we do have ``fromdataframe`` and similar methods and I don't think it will hurt to add this.

FrancescAlted:
IMO, NumPy is the default data container when you want to slice data from carray/ctable.  Adding a specific method for converting to NumPy means that we have two methods for doing the same, which can be confusing.

esc:
OK, this is true, but then let's add some more docs about accessing and programming ``carray`` internals. (at some point in the future).

PR #197, opened 2015-05-24
Fix using the auto_flush keyword in the ctable constructor

esc:
It wasn't possible to initialize a ctable with the auto_flush keyword at all.

Problem is: the **kwargs are being sent through to any new carrays and these
don't support this keyword argument.

There are multiple approaches to solving this problem:

a) Pop the auto_flush keyword from the **kwargs if it exists (current solution)

b) Implement auto_flush in the carray

c) Remove the auto_flush keyword entirely in favour of with statement.
   (Although this might not be enough to support all use-cases.)

### DIFF
--- a/bcolz/ctable.py
+++ b/bcolz/ctable.py
@@ -186,6 +186,8 @@ class ctable(object):
             self.auto_flush = kwargs.pop('auto_flush', True)
         else:
             self.auto_flush = False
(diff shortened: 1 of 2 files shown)


esc:
Gonna merge this up, maybe in future a better solution will be found.

PR #198, opened 2015-05-26
start making use of the chunks iterator

esc:
It became available with #153 here is how to use it.

### DIFF
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -1740,8 +1740,7 @@ cdef class carray:
         result = np.zeros(1, dtype=dtype)[0]
 
         nchunks = <npy_intp> cython.cdiv(self._nbytes, self._chunksize)
-        for nchunk from 0 <= nchunk < nchunks:
-            chunk_ = self.chunks[nchunk]
+        for chunk_ in self.chunks:
             if chunk_.isconstant:
                 result += chunk_.constant * self._chunklen
             elif self._dtype.type == np.bool_:


waylonflinn:
is this a good replacement for the following construction used in both `bdot` and `bquery`?

```python
for i in range(carray_.nchunks):
		chunk_ = matrix.chunks[i]
		chunk_._getitem(0, chunk_len, m_i.data)
```
Also, does it remove the need to separately process the leftover array?

If so, I will be overjoyed! Thanks!

CarstVaartjes:
I would be too! I have some questions though ->
- which of the implementations is it and how does it compare in speed? (see @FrancescElies his performance tests there)
- does it include leftovers? (because for me from a functional PoV, it really should)

I actually once tried something with a yield (cython supports that nowadays) and that worked really well too (and high performant as far as i can remember). if you guys want, I can make that example again?

CarstVaartjes:
@esc, @waylonflinn : See this as a thought experiment I had: https://github.com/CarstVaartjes/bcolz/commit/a06f24756e749d50d6a24881b571f223b483e73d

this would de-complicate bquery a LOT to be honest; i'm not sure if the chunks iterations as it's here would be usable (due to the questions above)

FrancescAlted:
Yes, I like your approach.  Just that the method should be called 'iterchunks()' no?

CarstVaartjes:
Ok! Will rename it, check if i can replace the leftover calculation with the leftover api merge from @esc and add some tests; expect a pull request this weekend ;)
I also have some questions on how wheretrue influences chunks, i'll take that to the mailing list!

esc:
This is for internal usage mainly to simplify a few arguments and doesn't handle the leftovers, and yes I like the ``iterchunks`` but it might be confused with ``iterblocks``. Anyway, I wouldn't worry too much since ``iterblocks`` kind of has the wrong name. "blocks" are the subunits of a "chunk". Looking forward to the PR.

esc:
Oh also, a typed version of this might be nice for cython, but lets tackle one thing at a time.

PR #200, opened 2015-05-26
remove compat layer and rename

esc:
IIRC before Python version 2.6 we still needed the compat layer, now we can
just go ahead and call everything PyBytes_* as it is suppoed to be in Python
3.x.


esc:
PR 200!

PR #202, opened 2015-05-27
avoid redundant memory copy in chunk.getudata

esc:
Micorbenchmarks to follow.

### DIFF
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -78,6 +78,7 @@ from definitions cimport (malloc,
                           PyBytes_AsString,
                           PyBytes_GET_SIZE,
                           PyBytes_FromStringAndSize,
+                          PyBytes_AS_STRING,
                           Py_BEGIN_ALLOW_THREADS,
                           Py_END_ALLOW_THREADS,
                           PyBuffer_FromMemory,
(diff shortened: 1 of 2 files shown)


esc:
With the optimization:

```
In [1]: import bcolz

In [2]: a = bcolz.arange(1e6, chunklen=500000)

In [3]: c = a.chunks[0]

In [4]: %timeit u = c.getudata()
1000 loops, best of 3: 704 µs per loop
```

esc:
```
In [1]: import bcolz

In [2]: a = bcolz.arange(1e6, chunklen=500000)

In [3]: c = a.chunks[0]

In [4]: %timeit u = c.getudata()
1000 loops, best of 3: 897 µs per loop
```

esc:
So, all in all not that significant. Though I am not sure you need to check for NULL return when calling ``PyBytes_FromStringAndSize`` in Cython.. Anyone know?

esc:
The timings are actually fairly volatile. I should imagine with larger chunks they may be more reliable.

FrancescAlted:
Well, I would not say that 20% is 'not that significant'.  So I would go ahead and merge this.

esc:
Please go ahead and press the button! :)

FrancescAlted:
I was commanded to "press the button", so here we go!

esc:
Sweeet!

FrancescAlted:
It was as good for you as it was for me?

esc:
Oh... it was amazing!!!

FrancescElies:
I see some sort of ZeroMQ contribution policy being applied here, nice!

esc:
Aw... aw... aw... yeah! https://www.youtube.com/watch?v=bWbJeUEKzc8

PR #208, opened 2015-06-21
fix bug for incorrect leftover value

waylonflinn:
In some rare cases using the standard mod operator in cython was causing the leftover value to
be incorrectly set to a non-zero value. This was causing segfaults in bdot (and probably bquery), from attempts to write outside the bounds of correctly sized result arrays in cython.

This example reproduced the bug in my tests.
```python
rows = 2 ** 21
columns = 200
matrix = np.random.random_integers(0, 120, size=(rows, columns))
chunk_len = 2 ** 2
bcarray = bdot.carray(matrix, chunklen=chunk_len, cparams=bdot.cparams(clevel=2))
```

See the 'Abstract' section of this document for details on the differences between standard mod and `cython.cmod`: [Enhancements Division](https://github.com/cython/cython/wiki/enhancements-division)

### DIFF
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -1212,7 +1212,7 @@ cdef class carray:
 
         # Update some counters
         calen = shape[0]  # the length ot the carray
-        self.leftover = (calen % chunklen) * self.atomsize
+        self.leftover = cython.cmod(calen, chunklen) * self.atomsize
         self._cbytes = cbytes
         self._nbytes = calen * self.atomsize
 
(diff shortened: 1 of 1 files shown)


FrancescAlted:
This looks good to me.  Thanks!

esc:
@FrancescAlted you forgot to add an entry for this in the RELEASE_NOTES, I'll take care of it now.

esc:
I have updated  de6208e97860c430957fce4565c30037175274ba

waylonflinn:
Thanks! Wasn't sure if I should!



—

esc:
You can if you want---and it would help; but in case you forget, I usually end up sweeping things together before a release.

PR #209, opened 2015-06-24
Carray private methods

FrancescElies:
Related to https://github.com/Blosc/bcolz/issues/22

I started marking some methods with a single underscore, would you prefer double underscore for that?

### DIFF
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -1005,7 +1005,7 @@ cdef class carray:
             _new = True
         elif rootdir is not None:
             meta_info = self._read_meta()
-            self.open_carray(*meta_info)
+            self._open_carray(*meta_info)
             _new = False
         else:
             raise ValueError(
@@ -1159,7 +1159,7 @@ cdef class carray:
         # and flush the data pending...
         self.flush()
 
-    def open_carray(self, shape, cparams, dtype, dflt,
+    def _open_carray(self, shape, cparams, dtype, dflt,
                     expectedlen, cbytes, chunklen, xchunks=None):
         """Open an existing array."""
         cdef ndarray lastchunkarr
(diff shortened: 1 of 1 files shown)


FrancescAlted:
No a single underscore is the standard de facto for private methods.  Is that ready for merging?

FrancescElies:
Nope, I just added only two of them

FrancescElies:
Now I could see the following methods:
```
In [1]: c = bz.carray([])

In [2]: c.
c.append          c.dtype           c.nchunks         c.shape
c.atomsize        c.flush           c.ndim            c.size
c.attrs           c.free_cachemem   c.next            c.sum
c.cbytes          c.itemsize        c.open_carray     c.trim
c.chunklen        c.iter            c.partitions      c.view
c.chunks          c.leftover_array  c.purge           c.where
c.copy            c.leftovers       c.reshape         c.wheretrue
c.cparams         c.len             c.resize          
c.create_carray   c.mode            c.rootdir         
c.dflt            c.nbytes          c.safe 
```

Does anybody see something that should not be there?

FrancescAlted:
Well, I propose you put your user's hat and propose some private functions.  If you don't find any candidate, then we have it already.  Actually I don't see that many functions, so I would say this is fine.

CarstVaartjes:
The c.open_carray doesn't make sense as a publicj, as when you create a carray it's opened at that time or a new one is created.

FrancescAlted:
Yeah, and create_carray is in the same bag.

FrancescElies:
Thanks guys, as a user the rest would seem ok to me

FrancescAlted:
Merging.

esc:
You forgot the changelog, entry, I'll ad it now.

esc:
:)

esc:
See: a2473f4082ccd925a83a86e85134fa29ca86ca48

FrancescElies:
cool beans! thanks :)

PR #212, opened 2015-06-29
Iterchunks

CarstVaartjes:
Hi,

I've created the iterchunks() function and added a test. Some comments:

- Atm it fails, because _getitem expects a string and not a ndarray(!?!) If anyone could tell me what I'm doing wrong? For details see: https://groups.google.com/forum/#!topic/bcolz/Fk6YpxD5toc 
- It still calculates the leftover elements itself, because nleftover has not been merged into master yet

Some tips to help me fix this, would be very much appreciated!
BR

Carst

### DIFF
--- a/bcolz/carray_ext.pxd
+++ b/bcolz/carray_ext.pxd
@@ -45,6 +45,7 @@ cdef class carray:
     cdef public object chunks
     cdef object _rootdir, datadir, metadir, _mode
     cdef object _attrs, iter_exhausted
+    cdef object iterchunks
     cdef ndarray iobuf, where_buf
     # For block cache
     cdef int idxcache
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -2337,7 +2337,7 @@ cdef class carray:
             cview = self.view()
         return cview._init_iter(start, stop, step, limit, skip)
 
-    def iterchunk(self):
+    def iterchunks(self):
         cdef:
             chunk chunk_
             Py_ssize_t chunklen, leftover_elements
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -2337,6 +2337,25 @@ cdef class carray:
             cview = self.view()
         return cview._init_iter(start, stop, step, limit, skip)
 
+    def iterchunk(self):
+        cdef:
+            chunk chunk_
+            Py_ssize_t chunklen, leftover_elements
+
+        chunklen = self.chunklen
+        # in-buffer ideally should be typed but that would need fusedtype?
+        in_buffer = np.empty(chunklen, dtype=self.dtype)
+
+        for i in range(self.nchunks):
+            chunk_ = self.chunks[i]
+            # decompress into in_buffer
+            chunk_._getitem(0, chunklen, in_buffer.data)
+            yield in_buffer
+
+        leftover_elements = cython.cdiv(self.leftover, self.atomsize)
+        if leftover_elements > 0:
+            yield self.leftover_array[:leftover_elements]
+
     def _init_iter(self, start, stop, step, limit, skip):
         self.reset_iter_sentinels()
         self.sss_mode = True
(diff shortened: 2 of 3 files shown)


FrancescElies:
Hi Carst,

I see Travis tests passing and Appveyor too (build x64 architecture is still pending, but x32 is passing)

CarstVaartjes:
It's so weird! Discussing it with @esc on Google Groups, not sure what is going wrong and if it's really an issue... (bquery should have issues with this too)

CarstVaartjes:
Hi,

a question: should i change the chunk ._getitem to expect a buffer as dest instead of a *char?

FrancescElies:
Not sure, tests are passing, are you still having problems on your environment?
You could try and see what happens, or maybe try casting

esc:
You could try and see if something has changed in numpy or cython by trying to compile with older versions, anaconda still has a bunch of older ones IIRC.

FrancescElies:
If I see that correctly `iterchunksTest` inherits only from `MayBeDiskTest` I guess travis is not executing that test, it should also inherit from `TestCase`

FrancescAlted:
That's correct.  Well spotted.

CarstVaartjes:
*hides head in shame*

FrancescAlted:
No no, that's not trivial and has bitten me a few times too.

2015-07-03 12:55 GMT+02:00 Carst Vaartjes <notifications@github.com>:

> *hides head in shame*
>
> —
> <https://github.com/Blosc/bcolz/pull/212#issuecomment-118315162>.
>



-- 
Francesc Alted

CarstVaartjes:
Thanks! So it gives the same error as I would expect; I will test what it does when I change the expected type from *char to buffer in cython

esc:
I'm not even sure buffer is available beyond 2.6

CarstVaartjes:
Nope, it gives horrible errors :) Ndarray does too (because of dependencies etc) hmmm :/ I still don't understand that bquery and bdot seem to have no issues but inside bcolz we do (likely some cython thing).
So we cannot cdef a buffer while ndarray.data is a buffer? does anyone know if there is something available from numpy for this? I have been googling at it but cannot seem to find anything

FrancescElies:
I am getting bit confused and having some questions, how does `iterchunks` differ from https://github.com/Blosc/bcolz/pull/153, is it performance? One should be able already to iterate over chunks, isn't it?
If the reason is performance wouldn't it be worth to apply those performance improvements on top of `__iter__` and `__next__` methods if possible?

If I understood correctly from a previous post https://github.com/Blosc/bcolz/pull/198#issuecomment-116316396 `iterchunks` shouldn't include leftovers right? Extensions writers still need to handle chunks and leftovers in two separate `for` loops.
On #153 we also had also a similar conversation, I understood that `iterblocks` could fill that requirement. 

Maybe would it be a better approach to use `PyBytes_AsString` instead of changing
the type declaration inside the function?

FrancescAlted:
I think the last note by @FrancescElies really applies.  @CarstVaartjes would `iterblocks` be enough for your needs?

CarstVaartjes:
Hi, the point (what I also made in the previous comments) was mostly that for me iterchunks explicitly should handle leftovers. The reason is that you do not want code to look like this:

    for array in iterchunks(carray_):
        do_stuff(a)

    if carray_.leftoverelements > 0:
        do_stuff(carray_.leftoverarray[:carray_.leftoverelements])

the second part is a bit ugly and really has no additional value. 
On the other hand, no clue how to solve the error we're getting here. it works perfectly fine in bquery where we cimport arrays... we added this there now:

    for array_len, array in iterchunks(carray_):
        do_stuff(a[:array_len])

and that works really well (and makes the code much more readable).
As i'm clueless in how to fix this with the cython weirdness popping up + a workaround of having it available in bquery, I propose to close this indeed. (just wanted to explain why you want more than just the chunks :)

FrancescAlted:
Ok, so I think we can safely close this. At any rate,  `iterblocks` already deals with the leftovers, so again, that maybe enough for your needs.

CarstVaartjes:
Thanks! I will check it out

PR #215, opened 2015-07-12
blosc http

FrancescElies:
can't connect to https://blosc.org, would http://blosc.org forwarding be ok?


FrancescAlted:
I don't know how to setup the https -> http forwarding rule, but this should do for now.  Thanks.

FrancescAlted:
BTW, why you closed the PR?

FrancescElies:
Not sure, I guess it was an accident I didn't even noticed that

PR #217, opened 2015-07-13
to_ndarray: proper conversion of numpy arrays when no dtype is specified

sdvillal:
This should fix a bug when creating *carrays* from non-contiguous numpy views without passing *dtype*. For example, this roundtrip fails with current master (bb6080e5285):

```python
import numpy as np
import bcolz
transposed_array = np.array([[0, 1, 2], [2, 1, 0]]).T
print(transposed_array)

[[0 2]
 [1 1]
 [2 0]]

print(bcolz.carray(transposed_array))

[[0 1]
 [2 2]
 [1 0]]
```

Same problem, but when passing a simple pandas dataframe:

```python
import pandas as pd
import bcolz
df = pd.DataFrame(data={
    'a': np.arange(3),
    'b': np.arange(3)[::-1]
})
print(df)

   a  b
0  0  2
1  1  1
2  2  0

print(bcolz.carray(df))

[[0 1]
 [2 2]
 [1 0]]
```


bcolz/tests/test_utils.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+########################################################################
```
esc: Please remove this file header, I guess.

sdvillal:
Done. But let me know if there is a better place for that test.

sdvillal:
And the test is wrong, will fix it now.

esc:
Hard to say, but I think it is the correct place for it. Not sure where the other stuff from utils is tested, but OK---let's wait for the tests to go green.

FrancescAlted:
Hmm, what about putting the new test in `test_carray.py`?  I think it fits quite well there.

sdvillal:
Done, hope style is consistent.

The original ```assert_remove_equal``` test was bogus, because it would always pass; but now I have added a full roundtrip test.

FrancescAlted:
Looks good to me.  Could you add an entry in RELEASE_NOTES.rst?  After that we would be all set.

sdvillal:
Done.

FrancescAlted:
Merged.  Thank you!

PR #218, opened 2015-07-14
do not assume that `np.array` does the right thing

sdvillal:
Continuation of #217. The fix there only works for transposed arrays, but it still fails to fix errors with, for example, pandas dataframes:

```python
import pandas as pd
import bcolz
df = pd.DataFrame(data={
    'a': np.arange(3),
    'b': np.arange(3)[::-1]
})
print(df)

   a  b
0  0  2
1  1  1
2  2  0

print(bcolz.carray(df))

[[0 1]
 [2 2]
 [1 0]]
```
This is the same example we had in #217, run with latest master (e260ca71bc) installed on latest anaconda. I'm working in a fix for this that will do the right thing (the first commit here is a step, but breaks other conversions).

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -32,7 +32,7 @@ Changes from 0.9.0 to 0.10.0
 - Various refactorings and cleanups. (#190 #198 #197 #199 #200)
 
 - Fix bug creating carrays from transposed arrays without explicit dtype.
-  (#217 @sdvillal)
+  (#217 #218 @sdvillal)
 
 
 Changes from 0.8.1 to 0.9.0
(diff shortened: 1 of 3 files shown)


sdvillal:
This should now be ready to merge.

FrancescAlted:
Well done!

esc:
Yeah, looks pretty perfect.

esc:
@sdvillal do you have more stuff lined up? Just asking because I'd get on with the 0.10.0 release otherwise.

sdvillal:
@esc No, that was all for the moment.

sdvillal:
Actually I just thought, for continuous integration to run all tests, we should install pandas in travis/appveyor (I guess using something like miniconda instead of pip to be nice to these services and speed-up builds).

FrancescAlted:
:+1:

PR #219, opened 2015-07-15
leverage setuptools for smoother install

waylonflinn:
automatically install cython and numpy with setuptools

* easier pip install
* enables use of airspeed velocity for benchmarking

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,11 +8,23 @@
 
 from __future__ import absolute_import
 
+# The minimum version of Cython required for generating extensions
+min_cython_version = '0.22'
+# The minimum version of NumPy required
+min_numpy_version = '1.7'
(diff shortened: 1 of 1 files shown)


waylonflinn:
Was hoping to get this in before 0.10.0, as it enables me to use airspeed velocity to benchmark bdot.

Here's some additional references for the weird parts of the code:

* [setuptools Distribution hack](http://stackoverflow.com/questions/12060925/best-way-to-share-code-across-several-setup-py-scripts/12061891#12061891)
* [`__NUMPY_SETUP__` hack](http://stackoverflow.com/questions/21605927/why-doesnt-setup-requires-work-properly-for-numpy)

FrancescAlted:
Yep, looks good to me.  Thanks!

PR #225, opened 2015-07-17
Travis using conda

FrancescElies:
Installs numpy pandas cython for every environment, mock and unittest2 depending on environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,30 @@
 sudo: false
 language: python
+
 python:
 - 2.6
 - 2.7
 - 3.4
+
 install:
-- pip install -r requirements.txt
-- pip install cython
-- if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install unittest2 --upgrade; fi
-- python setup.py build_ext --inplace
+  # Install conda
+  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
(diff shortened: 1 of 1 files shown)


esc:
I didn't know 2.6 was avail through conda, nice.

bcolz/tests/test_ctable.py
```diff
@@ -19,7 +19,10 @@
 from bcolz.py2help import xrange, PY2
 import pickle
 import os
-from mock import Mock
```
esc: Compatability code should live in `bcolz/py2help`.

FrancescElies: I didn't notice we had something for that, I'll move those changes to a new PR, thanks!

FrancescElies:
I would say travis is now ready to use conda

FrancescAlted:
Looks good.  Merging.

FrancescAlted:
Do you have an estimation on how fast this is respect the previous deployment?

FrancescAlted:
Just checked; apparently Travis completes checks 2x faster, which is cool.

FrancescElies:
nice :), appveyor speed up was also quite good, before it took from 6 to 9 minutes for each point in the test-matrix, now it takes between one and two minutes for each point in the test-matrix

FrancescAlted:
Oh, that's much better than I realized.

esc:
Well done!

Issue #229, opened 2015-07-22
ctable.append fails when argument is single row from itself


waylonflinn:
The append method cannot accept the return value of it's own accessor, for a single row.

This fails:

```python
example_table.append(example_table[0])
```

while this works:

```python
example_table.append(example_table[0:1])
```

The first one fails because the type of the return value is `numpy.void` instead of `numpy.ndarray`, 
the check is on this line in ctable: [ctable.py#L354](https://github.com/Blosc/bcolz/blob/bb148ffaf9a035d71cbd004d07d59ba66800e019/bcolz/ctable.py#L354)

PR #231, opened 2015-07-26
added support for appending np.void in ctables. fixes ticket #229

eumiro:
Added a new test for this enhancement.


FrancescAlted:
This looks good, but it would be nice if you can add an entry in the RELEASE_NOTES.txt file.  Could you please do that?

FrancescAlted:
Merged. Thank you very much for a nice contribution!

PR #232, opened 2015-07-26
using all() to check types for a list of objects

eumiro:
Using all() instead of comparing a list of objects to a list of types allows more flexible checks and stops as soon as False is found.


FrancescAlted:
Looks good to me.  Thanks!

esc:
Does this still need a RELEASE_NOTES entry?

FrancescAlted:
Well, this was done during a mini-sprint at EuroPython, and I suggested that as this was a quite minor change, and that an entry in release notes was not needed.  But feel free to mention that if you find it is worth to.

PR #233, opened 2015-07-27
Fe/ci/conda.recipe

FrancescElies:
This PR offers the follwoing for CI travis and appveyor:
- Instead of managing dependencies for both CI systems separately, uses `meta.yaml` as a centralised file.
- Fix: `CONDA_NPY: "1.9"` to `CONDA_NPY: "19"`
- Move Package to make the CI aware of created artifacts, see e.g. https://ci.appveyor.com/project/FrancescElies/bcolz/build/job/gjfo47pewpekocwo/artifacts
- Modifies `conda.recipe/bld.bat ` and `conda.recipe/build.sh`, I doubt if I did the right thing here, but I could not make it work without modifying them, would you be OK with these modifications?

Based on https://github.com/rmcgibbo/python-appveyor-conda-example/

Thanks in advance for the time you invest reviewing this.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,6 +1,10 @@
 package:
   name: bcolz 
 
+source:
+    git_rev: v1.7.0
+    git_url: https://github.com/pypa/setuptools_scm.git
+
 requirements:
   build:
     - python
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -11,7 +11,7 @@ build:
 requirements:
   build:
     - python
-    - setuptools_scm 
+    - setuptools-scm 
     - cython
     - numpy
 
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -11,7 +11,7 @@ build:
 requirements:
   build:
     - python
-    - setuptools-scm 
+    - setuptools_scm 
     - cython
     - numpy
 
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,10 +1,6 @@
 package:
   name: bcolz 
 
-source:
-    git_rev: v1.7.0
-    git_url: https://github.com/pypa/setuptools_scm.git
-
 build:
     preserve_egg_dir: True
 
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,16 +1,11 @@
 package:
   name: bcolz 
-  version: "0.7.2"
-
-build:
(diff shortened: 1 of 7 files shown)


continuous-integration/travis/install.sh
```diff
@@ -0,0 +1,9 @@
+MINICONDA_URL="http://repo.continuum.io/miniconda"
+MINICONDA_FILE="Miniconda-3.5.5-Linux-x86_64.sh"
```
FrancescAlted: There would a way to do not have to specify a version of Miniconda?

FrancescAlted:
Other than the comment above, this looks pretty damn good.  Thanks!

continuous-integration/travis/install.sh
FrancescElies: yes good point, we could use latest, I'll change that

FrancescElies:
I think I saw tests pass, now failing, ups
Anyway maybe it needs a rebase too, not sure what's going on.

FrancescAlted:
Strange.  I would swear that I have seen the tests passing too :/

FrancescElies:
Seems like `setuptools_scm` needs to be added as a build time dependency, but cannot be found ...

FrancescAlted:
This should be related with PR #239

FrancescElies:
Not yet possible to add it as a build requirement http://stackoverflow.com/questions/25610843/conda-build-requirement-add-package-from-specific-channel.

For the moment I'll put the bug fixes in a different PR

FrancescElies:
Maybe adding a channel containing `setuptools_scm` would do the trick, not sure if it's worth the effort of maintaining a conda channel only for this.
Once I try https://binstar.org/elies/setuptools_scm out I'll report it back

continuous-integration/travis/install.sh
FrancescElies: I proposed this change for the original project too https://github.com/rmcgibbo/python-appveyor-conda-example/pull/2

conda.recipe/meta.yaml
```diff
@@ -1,16 +1,19 @@
...
+    git_url: https://github.com/pypa/setuptools_scm.git
 
 build:
-    number: {{environ.get('BINSTAR_BUILD', 1)}}
+    preserve_egg_dir: True
 
 requirements:
   build:
     - python
+    - setuptools_scm
```
mindw: That should be setuptools-scm

FrancescElies: sharp, thanks!
No luck though, I tried in  6ccc459 but still failing, I think this package might be missing in the default channel

FrancescElies:
Bug-fixes were already merged in different PR, remaining features are not working, closing this.

PR #247, opened 2015-08-29
Fe/ci/conda.recipe v2

FrancescElies:
Bugfixes from  #233

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,39 +11,39 @@ environment:
       PYTHON_VERSION: "2.6"
       PYTHON_ARCH: "32"
       CONDA_PY: "26"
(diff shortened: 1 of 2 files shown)


FrancescAlted:
I don't think I understand the workflow here.  This overrides #233 or both need to be applied?

FrancescElies:
oh sorry for the poor explanation, I thought in the following days I would have a look at cython documentation issue and separate the items not affected by `setuptools_scm` issue so that we could get them merged in the mean while, would you be ok with this?.
- fix CONDA_NPY environmental variable
- Use Miniconda latest

FrancescAlted:
Ah, I think that makes sense now.  Will merge this now.  Your effort is very appreciated!

PR #246, opened 2015-08-28
Fe/cython coverage

FrancescElies:
https://github.com/Blosc/bcolz/pull/227 continuation

PR #227, opened 2015-07-20
config for cython coverage

esc:
Doesn't seem to work yet.

PR #248, opened 2015-08-31
Fe/cython coverage gcov

FrancescElies:
This is a different try-out using gcov approach rather than cython built in coverage analysis used in #246 #227

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[run]
+source = bcolz
+
+[report]
+omit =
+    */python?.?/*
+    *bcolz/tests/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
   - source activate bcolz_env
   - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then conda install unittest2 mock; fi
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then conda install mock; fi
-  - pip install coveralls
+  - pip install cpp-coveralls
   
   # Build bcolz
   - python setup.py build_ext --inplace
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,10 @@ install:
   - conda update conda
 
   # Install dependencies
-  - conda create -n bcolz_env python=$TRAVIS_PYTHON_VERSION numpy pandas cython nose
+  - conda create -n bcolz_env python=$TRAVIS_PYTHON_VERSION numpy pandas cython nose coverage
   - source activate bcolz_env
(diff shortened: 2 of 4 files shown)


FrancescElies:
at least c and h files seem to be listed by coveralls, but nosetests results disappeared from the results list and to see which lies are covered the c source needs to be on github https://coveralls.io/builds/3441555/source?filename=bcolz%2Fcarray_ext.c

FrancescAlted:
Well, adding the .c sources is not necessarily a bad practice, rather the contrary.  Here it is a function that might be helpful for that purpose:

https://github.com/PyTables/PyTables/blob/develop/setup.py#L651

and then add the .c files in the MANIFEST.in like in:

https://github.com/PyTables/PyTables/blob/develop/MANIFEST.in#L5

Would you mind to file another PR with that?

FrancescElies:
I just opened a new PR for that, but this one is not yet ready to merge though, the coverage of all python doesn't appear in coveralls anymore

FrancescAlted:
That's right, '-fprofile-arcs -ftest-coverage -lgcov' will only have to be active during Travis runs, so it does not matter whether they have a performance impact or not.

FrancescElies:
Still very patchy and /bcolz/carray_ext.c is getting more than a 100% coverage

FrancescElies:
Now seems python and c files to be reported at the same time

FrancescAlted:
This looks pretty good to me.  Should we wait until carray_ext.c is in the repo before merging this, right?

FrancescElies:
Now we get 65% coverage on the carray_ext.c, once appveyor finishes I guess we could actually merge it, you'll see the percentage on coveralls but once you click on the file you won't be able to see which lines on `carray_ext.c` are covered because this files is not on github

FrancescElies:
From what I see https://github.com/PyTables/PyTables/blob/develop/setup.py#L651 will get the c sources into the tar ball of the package right?

Coveralls needs the c source to be on github but keeping the pyx and the c file tracked by git is something we would like to avoid.

FrancescAlted:
Yeah, let's start by merging this one then.  Thanks.

FrancescAlted:
That's a good point: the procedure above is for putting the C sources in the tarball.  Now, that I think more about this, I don't really like the possibility to add the Cython C sources in the repo because if you change a single line in Cython, the diff in the C sources can be between 10 and 10 zillion lines, so not fun when you are reading diffs in a browser :(

Because of this, my vote would be -0 for adding the Cython C sources into the repo.  We should be happy to see total coverage then and not the line detail (which is pretty deceiving, but anyway).

FrancescElies:
I agree, the best would be to get #246 working, and see pyx files coverage directly, but no luck yet

Issue #238, opened 2015-08-20
Make bcolz.carray subclassing  simpler


mpage38:
Currently, subclassing bcolz.carray is difficult, because in bcolz.ctable (and maybe in other classes), there exist tests like if type(...) == bcolz.carray. It means that ctable must also be subclassed with the purpose of overriding these methods and replacing those tests with: if issubclass(..., bcolz.carray).
Could these changes be done directly in bcolz library ?

FrancescAlted:
Yes, that's a good point.  Pull requests are encouraged ;)

mpage38:
One word to back-up my feature request: you kept the bcolz API quite simple and I think that it is a good idea ... as long as it is possible to subclass carray and ctable. If not, it is necessary to use composition to build something on top of bcolz, and this can easily become clumsy.

FrancescAlted:
I think this should be fixed in latest releases.  If not, please reopen this ticket and add a minimal code showing the issue.

PR #254, opened 2015-09-22
issubclass carray

FrancescElies:
Related to https://github.com/Blosc/bcolz/issues/238
@mpage38 would this be enough for your needs?


FrancescAlted:
I don't think this is what @mpage38 is requesting in #238.  He asked for replacing `type(object) == bcolz.carray` by something like `isinstance(object, bcolz.carray)`.

FrancescElies:
closing this

Issue #255, opened 2015-09-27
empty bcolz ctable


FrancescElies:
Coming from the mailing list, Recommended way to create an empty bcolz ctable? https://groups.google.com/forum/#!topic/bcolz/oBsiCv7qcS4

At the moment is possible to do the following, which creates an empty carray
```
In [7]: bcolz.zeros(0, dtype="i4")
Out[7]: 
carray((0,), int32)
  nbytes: 0; cbytes: 16.00 KB; ratio: 0.00
  cparams := cparams(clevel=5, shuffle=True, cname='blosclz')
[]
```

It would be nice to make it possible for empty ctables too (not yet possible though):
```
In [7]: bcolz.zeros(0, dtype="i4,f8")
Out[9]: 
ctable((0,), [('f0', '<i8'), ('f1', '<f8')])
  nbytes: 0; cbytes: 32.00 KB; ratio: 0.00
  cparams := cparams(clevel=5, shuffle=True, cname='blosclz')
[]

```

PR #256, opened 2015-09-28
Fe/feature/ctable zeros

FrancescElies:
Starting https://github.com/Blosc/bcolz/issues/255

### DIFF
--- a/bcolz/toplevel.py
+++ b/bcolz/toplevel.py
@@ -294,7 +294,7 @@ def fill(shape, dflt=None, dtype=np.float, **kwargs):
             os.mkdir(base_rootdir)
         for name,col_dype in dtype.descr:
             if base_rootdir is not None:
-                kwargs['rootdir'] = "{}/{}".format(base_rootdir, name)
+                kwargs['rootdir'] = "{0}/{1}".format(base_rootdir, name)
             dflt = np.zeros((), dtype=col_dype)
             ca = bcolz.carray([], dtype=col_dype, dflt=dflt, 
                     expectedlen=expectedlen, **kwargs)
--- a/bcolz/toplevel.py
+++ b/bcolz/toplevel.py
@@ -232,7 +232,7 @@ def fill(shape, dflt=None, dtype=np.float, **kwargs):
     """
     fill(shape, dtype=float, dflt=None, **kwargs)
 
-    Return a new carray object of given shape and type, filled with `dflt`.
+    Return a new carray or ctable object of given shape and type, filled with `dflt`.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


bcolz/toplevel.py
```diff
@@ -259,6 +259,21 @@ def fill(shape, dflt=None, dtype=np.float, **kwargs):
 
     """
 
+    def fill_helper(obj, dtype=None, length=None):
+        """Helper function to fill a carray with default values"""
+        assert issubclass(obj.__class__, bcolz.carray)
```
FrancescAlted: What's the difference between this and the more direct `assert isinstance(obj, bcolz.carray)`?

bcolz/toplevel.py
```diff
@@ -273,22 +288,28 @@ def fill(shape, dflt=None, dtype=np.float, **kwargs):
...
     if dtype.kind == "V" and dtype.shape == ():
-        raise ValueError("fill does not support ctables objects")
-    obj = bcolz.carray([], dtype=dtype, dflt=dflt, expectedlen=expectedlen,
+        list_carrays = []
+        base_rootdir = kwargs.pop('rootdir', None)
+        if base_rootdir is not None:
+            os.mkdir(base_rootdir)
+        for name,col_dype in dtype.descr:
+            if base_rootdir is not None:
+                kwargs['rootdir'] = "{0}/{1}".format(base_rootdir, name)
```
FrancescAlted: Is `{0}/{1}` portable for windows or we would need to use `os.join()`?

bcolz/toplevel.py
```diff
@@ -273,22 +288,28 @@ def fill(shape, dflt=None, dtype=np.float, **kwargs):
     # Create the container
     expectedlen = kwargs.pop("expectedlen", length)
     if dtype.kind == "V" and dtype.shape == ():
-        raise ValueError("fill does not support ctables objects")
-    obj = bcolz.carray([], dtype=dtype, dflt=dflt, expectedlen=expectedlen,
+        list_carrays = []
+        base_rootdir = kwargs.pop('rootdir', None)
+        if base_rootdir is not None:
+            os.mkdir(base_rootdir)
+        for name,col_dype in dtype.descr:
```
FrancescAlted: PEP8: need space between `name` and `col_dype`.  Also, `col_dype` -> `col_dtype`.

bcolz/toplevel.py
FrancescElies: None, I guess I was somehow subconsciously influenced by the topic subclassing carray https://github.com/Blosc/bcolz/issues/238

bcolz/toplevel.py
FrancescElies: I'll run pep8 for the whole function

bcolz/toplevel.py
FrancescElies: Not sure, but I'm afraid you'r right, anyway, I think I'll remove that. When creating those carrays, I need to create the base dir too; once created in the end I try to create the ctable container, it abiously complains because the directory already exists.

I think I'll create the carrays in memory if that's fine (unless it is a bad idea), for the sake of an easier review I'll make a new commit for that

FrancescElies: I made those changes in 69d9763, let me know if you feel that would be ok

FrancescAlted:
This is looking pretty good.  It would be really nice if you could add some small mention to new functionality in the tutorials and an small entry in RELEASE_NOTES.

FrancescElies:
Sure no problem, I'll do that in the following commits

FrancescAlted:
Merging this.

FrancescAlted:
Uh, could you please rebase this?  Thanks.

FrancescAlted:
Added in https://github.com/Blosc/bcolz/commit/f2454f1bec573f198a2bc1c50370193364a45f5f.  Closing this.

FrancescElies:
No need to rebase I guess, I see  f2454f1 is on master

FrancescAlted:
That's correct, this is in master.

Issue #257, opened 2015-09-29
Check documentation examples


FrancescElies:
Some examples in the documentation are out of date https://github.com/Blosc/bcolz/blob/master/doc/tutorial.rst

In the following example one would not expect a void type in the ctable
```
>>> N = 100*1000
>>> ct = bcolz.fromiter(((i,i*i) for i in xrange(N)), dtype="i4,f8", count=N)
>>> ct
ctable((100000,), |V12) nbytes: 1.14 MB; cbytes: 283.27 KB; ratio: 4.14
  cparams := cparams(clevel=5, shuffle=True, cname='blosclz')
[(0, 0.0), (1, 1.0), (2, 4.0), ...,
 (99997, 9999400009.0), (99998, 9999600004.0), (99999, 9999800001.0)]
```
Instead one should get
```
ctable((100000,), [('f0', '<i4'), ('f1', '<f8')])
  nbytes: 1.14 MB; cbytes: 675.24 KB; ratio: 1.74
  cparams := cparams(clevel=5, shuffle=True, cname='blosclz')
[(0, 0.0) (1, 1.0) (2, 4.0) ..., (99997, 9999400009.0)
 (99998, 9999600004.0) (99999, 9999800001.0)]
```

FrancescElies:
closing, tutorial examples being moved to ipython notebook #261

PR #258, opened 2015-10-01
Fe/documentation/review

FrancescElies:
Start #257 

For easier review I'll write the questions I have inside the commit message, this way they will be self contained and we won't have to make back references to certain lines in the files, would you be ok with that? e.g. ac85489.

### DIFF
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -110,10 +110,13 @@ method.
 
 For example, if `b` is a carray with 10 million elements::
 
+  >>> b = bcolz.arange(10*1e6)
   >>> b
-  carray((10000000,), float64)  nbytes: 80000000; cbytes: 2691722; ratio: 29.72
+  carray((10000000,), float64)
+    nbytes: 76.29 MB; cbytes: 2.94 MB; ratio: 25.92
     cparams := cparams(clevel=5, shuffle=True, cname='blosclz')
-  [0.0, 1.0, 2.0... 9999997.0, 9999998.0, 9999999.0]
+  [  0.00000000e+00   1.00000000e+00   2.00000000e+00 ...,   9.99999700e+06
+     9.99999800e+06   9.99999900e+06]
 
 it can be enlarged by 10 elements with::
 
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -305,11 +305,11 @@ powerful) way is by using its set of iterators::
 The iterator also has support for looking into slices of the array::
 
   >>> %time sum(v for v in b.iter(start=2, stop=20, step=3) if v < 10)
-  CPU times: user 0.00 s, sys: 0.00 s, total: 0.00 s
-  Wall time: 0.00 s
+  CPU times: user 1.6 ms, sys: 560 µs, total: 2.16 ms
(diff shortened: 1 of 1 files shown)


FrancescElies:
specially b082ea5 and e80bb8c would require a second pair of eyes.
The rest of the commits are mostly small differences in cbytes due to c-blosc version

FrancescElies:
since it strongly depends on platform, shall I leave out those benchmark updates?
So let's say only change lines where it is obviously an old version of bcolz, stuff like cbytes or mode=w

doc/tutorial.rst
```diff
@@ -361,16 +383,16 @@ However, you must be aware that modifying a carray is expensive::
   >>> a = np.arange(1e7)
   >>> b = bcolz.carray(a)
   >>> %timeit a[2] = 3
-  10000000 loops, best of 3: 101 ns per loop
+  10000000 loops, best of 3: 94.4 ns per loop
   >>> %timeit b[2] = 3
-  10000 loops, best of 3: 161 µs per loop  # 1600x slower than NumPy
+  1000 loops, best of 3: 274 µs per loop # 2900x slower than NumPy
 
 although modifying values in latest chunk is somewhat more cheaper::
```
FrancescAlted: "somewhat more cheaper" -> "somewhat cheaper"

FrancescAlted:
No, rather the contrary.  Keeping the benchmarks updated with more modern computers is good, IMO.  So no need to change your current commits.

FrancescAlted:
I just had a shower and came to the realization that converting the tutorial into a Jupyter notebook can be the solution that we are looking for.  Jupyter notebooks can be rendered automatically in github since some months ago, and regenerating the output from it prier each release should be very easy.  One can even add some plots for some benchmarks too so as to make the read not that boring.  Ah, the showers :)

FrancescElies:
That sounds an appealing idea, something similar to bquery notebook http://nbviewer.ipython.org/github/visualfabriq/bquery/blob/ipynb_bench/bquery/benchmarks/bench_groupby.ipynb ?

Going to the swimming pool, let's see if there I can get some ideas too :)

FrancescElies:
I just found the github link https://github.com/visualfabriq/bquery/blob/ipynb_bench/bquery/benchmarks/bench_groupby.ipynb, would be this what you mean?

FrancescElies:
It will sound familiar to you since this was strongly based on the following notebooks
Blaze - http://nbviewer.ipython.org/url/blaze.pydata.org/notebooks/timings-bcolz.ipynb
Bcolz - http://nbviewer.ipython.org/github/Blosc/movielens-bench/blob/master/querying-ep14.ipynb

FrancescAlted:
Yes, that's exactly what I meant.  In fact, I did something along these lines for past EuroPython:

https://github.com/FrancescAlted/EuroPython2015/blob/master/0-Intro.ipynb

Initially I wanted all the docs in a Sphinx format, but now I realize that an IPython notebook is probably a much better fit for the tutorial chapters.

FrancescElies:
from my side +1 for the notebook. Shall we close this one?

FrancescAlted:
No, let's merge it and start work on a replacement.

PR #259, opened 2015-10-02
Add a logo! It's fun!

waylonflinn:
Whipped this up for `bcolz`, to match the one for `bdot`. It's totally simple pixel art.

Let me know if you like it, and I'll see if Carst wants one for `bquery`.


FrancescAlted:
Thanks for the logo suggestion!  With this pixelation, it looks to me like we are back to the times of Super Mario :)  I don't dislike it actually.  May I ask what it represents the three lines before the 'b'?  Columns?  And if so, why one of them is reversed?

waylonflinn:
Yeah, it's definitely got the Mario pixel-style. The things before the b _are_ columns. Reversing one just seemed to give it some character.

I'm hoping we can get bquery on pip soon, and I thought it might be nice to create some kind of common visual branding.

FrancescAlted:
I strongly agree that we need some visual branding, so let's start with this now and refine it in the future.  Thanks again!

PR #253, opened 2015-09-22
Fe/asv

FrancescElies:
In case still interested, maybe first steps towards https://github.com/Blosc/bcolz/issues/116?

Modifications:
- Basic config file  `asv.conf.json`
- Added a small guide at `bench/README_asv.md`.
- Benchmarks would need some extra boilerplate (see  `bench/arange.py`, `bench/column_iter.py`) and _maybe some renaming asv does not like dashes (-)_.
- Modified benchmarks respect `run.sh` (they still could be ran with this script).
- Benchmarks for Python 3 not working, not sure why.

Running it:
- For the moment to get it running only adapted scripts should show up in the `bench` folder (run first `rm -f bench/*` and `git checkout bench/arange.py bench/column_iter.py bench/__init__.py`)
- Preview available at http://francescelies.github.io/bcolz/ (only some points can be seen, but if desired one could ran it for every commit in the repository, would take a while though).

PR #260, opened 2015-10-05
fix: asv gets confused with `bcolz` project name

FrancescElies:
@FrancescAlted this should fix https://github.com/Blosc/bcolz/pull/253#issuecomment-145458834
Could you please give it a try

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -4,7 +4,7 @@
     "version": 1,
 
     // The name of the project being benchmarked
-    "project": "bcolz",
+    "project": "_bcolz", // underscore is intentional
 
     // The project's homepage
     "project_url": "https://github.com/Blosc/bcolz",


FrancescAlted:
Woohoo, it works now!  Now I am supposed to find a machine for running the benchmarks and showing the plots right?

FrancescAlted:
Also, any hint on why the original 'bcolz' was causing problems?  Just curious...

FrancescElies:
Honestly, no idea, I saw something with git cloning and at the end of the path appeared three times bcolz, that was a bit strange, I also remembered also changing the name of the project, to bcolz, which to me seemed a sensible name to choose :). 

My guess is that, git tries to create some folder named bcolz at the root level where the conf files is, since here bcolz already exists we get an error.

Better than renaming the underscore would be to move the conf file inside the bench_as folder, ask also creates other directories in the root dir which are a bit annoying, this way everything related to ask would be inside that folder.

FrancescElies:
I did some extra tests, the problem is, as suspected, a folder with the name of the project is being created at the same level of the configuration file if you merge #262 we can have it with the right name

FrancescElies:
Exactly, now you only need a machine and run it

FrancescElies:
but now you need first to change directory `cd bench_as` and then run the asv commands

PR #261, opened 2015-10-06
tutorial notebook

FrancescElies:
Migrating tutorial to python notebook.

To see the state of the notebook please visit:
https://github.com/FrancescElies/bcolz/blob/fe/doc/ipython_notebook_tutorial/doc/tutorials.ipynb


FrancescElies:
Not sure why appveyor is now failing it seems like suddenly on windows conda is installing dev version of numpy, https://github.com/Blosc/bcolz/pull/264 will force to install numpy 1.9

FrancescAlted:
Hey @FrancescElies this is looking great.  Thanks for your efforts.  Some comments:

I would really use the idiom:

```
c = bcolz.carray(a, rootdir='mydir')
```

instead of the more complex (and quite strange, specially for beginners):

```
>>> with bcolz.carray(a, rootdir='mydir') as _:  # for on-disk storage
...     c = _  # keep the reference to the Python object
```

Also, in the memory profiling tutorial (btw, very nice addition!), the plot for the iterator of the on-disk carray takes less memory than in-memory.  I think this is because the carray object needs to initialize some internal buffers (maybe some Python ones).  The fact that the on-disk carray is not taking the same memory perhaps is due to the fact that the previous buffers (either by bcolz or Python) can be reused.  It is probably worth mentioning that (but without saying that this is *the* correct explanation, because I am not sure; with memory one never knows ;).

Finally, I see that you are using the readthedocs site for the rest of the documentation, which is great.  Could you please add some directions in `RELEASING.rst` so that it would be easy to check them out for the poor release manager?

There could be other things, but perhaps fixing the above and merging would be fine for now.  Then, we can go on and continue refining.

FrancescElies:
Hi,

the plot for the iterator of the on-disk vs. memory is something that really surprised me and wanted to ask you about, I was really wondering what was going on, but you already saw that :).

As you might have already seen the completely new parts are the intro to the tutorials and the memory profiling carray, these are the parts that probably will require most of the corrections.

I'll see if I have some time in these days to fix the things you mentioned.

FrancescElies:
I changed the with statement, but I still make reference to it if you feel it would be better to remove it completely, please tell me.

This is a minor issue ,but I noticed internal references (Index and got to index) don't work if you look at the tutorial on github, but they do work if you have a look to the notebook with nbviewer, maybe you would like to reference links to nbviewer instead of github?

About the carray memory profiling I was still curious and did a small test, at this moment first we measure in-mem iter and then on-disk iter, so, if you change the order in which you measure the iterators (first on-disk), then the on-disk consumes more memory.
To me it seems like, your intuition was right, some sort of buffer is being reused.

About readthedocs and RELEASING.rst, I think it could be integrated with github and therefore no extra efforts would be needed (I never did this before, so I'm not completely sure).
At this moment it seems readthedocs points to valentins's fork https://github.com/Blosc/bcolz/issues/263

FrancescElies:
Before merging I could pack everything in a single commit if you like

FrancescElies:
Hi,

I think I'll replace the links from readthedocs to http://bcolz.blosc.org, somehow RTD has problems compiling https://readthedocs.org/projects/bcolz-fork-elies/builds/3406732/

FrancescAlted:
Hmm, strange, this happens after 0.11.4 was released yesterday?  If so, the problem should be in:

https://github.com/Blosc/bcolz/commit/597b2eca98bd616aebe828c442a2ccda58da7093#diff-2eeaed663bd0d25b7e608891384b7298L53

But this was my fix for #266 .  Bummer.

FrancescAlted:
Also, I don't understand why on earth RTD needs a build of the package (but I never used it, so I might be missing an important point here :/)

FrancescAlted:
At any rate, if RTD does not work for use I am +1 for the change of references to bcolz.blosc.org.  Furthermore I see that links to bcolz.blosc.org work perfectly well, right?
Finally, I don't see you changed the `with` contexts (you forgot a commit maybe?).

FrancescElies:
I think 597b2ec is unrelated to RTD.

FrancescElies:
Yep right I think i forgot to commit :/

FrancescElies:
With references to bcolz.blosc.org would be no extra actions needed in the RELEASING.rst, right?

FrancescElies:
Links to blosc.bcolz.org work fine, only the inner links don't work in github

FrancescAlted:
Ok, so if only the internal links are a problem, I would get rid of them. This is just a small price to pay for the convenience of having the notebook rendered directly inside github (i.e. no need to update the gist manually).

FrancescElies:
Well not really big a problem, when you click on them, no action is performed but they are quite handy when you open the notebook locally and want to edit/find certain section.
Let me know if you would still like them inside or not.

I just added https://github.com/Blosc/bcolz/commit/44e1e95b6427917d420625e32c84bd5480daf8a9 adding a comment you posted about mem profiling

FrancescAlted:
That's correct, if we continue to use bcolz.blosc.org then no further modifications in RELEASING.txt is needed.

FrancescAlted:
I think this is good enough for merging.  As soon as I find some time, I'll do another glance at that.  Thank you!

FrancescAlted:
Just for reference, in commit bfb9414 I have been doing a quick glance over your new `tutorial_carray_memory_profile.ipynb` notebook on my Ubuntu laptop, and I am not seeing the 4 MB that the iterator takes in your machine (Mac?).  In fact, it is *releasing* memory (yeah, a bit crazy).  So I decided to be quite less specific on detailing why the system is booking (or not :) memory and just remarking that bcolz iterators are both fast and memory-efficient.

FrancescElies:
Cool! thanks for your review, sure :), Yes tested it on a mac, atm I don't have much free time, but I could try it on a linux machine I'll let you know if I find some time to test this again, let's see if I get some memory released too :).

PR #262, opened 2015-10-07
Fe/asv/move files

FrancescElies:
This is a much better solution rather than #260, scipy uses exactly this structure


FrancescAlted:
Looks good to me.  Thanks!

PR #271, opened 2015-10-20
np 1.10 compat: replace "f16" dtype with np.longdouble

msarahan:
"Works for me"


FrancescAlted:
LGTM.  Would you mind adding an entry in RELEASE_NOTES.rst?

FrancescAlted:
I am thinking that after this PR a release would be worth the effort.

FrancescAlted:
Thanks!

Issue #272, opened 2015-10-20
Can't use unicode string as where clause


BrenBarn:
On Python 2.7 passing a unicode string as the query to `where` fails with an error like:

```
Traceback (most recent call last):
  File "<pyshell#5>", line 1, in <module>
    bcBasic.where("STATE=='California'", outcols=["nrow__", "GISJOIN"])
  File "c:\users\brenbarn\documents\python\extensions\bcolz\bcolz\ctable.py", line 852, in where
    "only boolean expressions or arrays are supported")
ValueError: only boolean expressions or arrays are supported
```

But using a bytestring works.

This is probably because of [this code](https://github.com/Blosc/bcolz/blob/0a061e2cf3f77c3a1493f26aada8bb56c6785b5f/bcolz/ctable.py#L845) in ctable.py, which explicitly checks if the query type `is str`.

I use Python 2.7 with `from __future__ import unicode_literals` and it is annoying to have to specify bytestrings.  I don't know if there are deeper reasons why unicode won't work (e.g., when passing to numexpr), but it would be nice.  If there are such deeper reasons, one possibility is to try encoding to ASCII (or whatever numexpr needs) and then raise a more informative error if it fails (e.g., "unicode string can't be used as query").  The current error message is confusing.

FrancescElies:
I can't tell exactly the reason why, but I see numexpr doesn't support it yet https://github.com/pydata/numexpr/issues/150, you could use python virtual machine as a workaround (won't be that fast but should work).
`bcolz.defaults.eval_vm = 'python'`
After executing this on your python session your where statement should work with unicode.

BrenBarn:
Well, a simpler workaround is to just use a bytestring (if entering a literal query, like in my example) or encode the query to ASCII (if you have it in a variable).  This imposes no speed penalty.  But my point is it would be better to have some of these workarounds built in.

FrancescAlted:
Does 668c4c9 help?

BrenBarn:
That doesn't seem to fix it, because that commit just adds handling of bytestrings.  Bytestrings already work in Python 2.  The problem is that in Python 2, `str` is a bytestring, and `unicode` is a separate type.  So the checks in various places that do things like `if isinstance(blah, str)` fail if `blah` is actually a unicode object.

PR #274, opened 2015-11-05
Py2strings

BrenBarn:
This fixes up string-type handling in a few places.  Bcolz has a py2helper module for aiding Python 2/3 compatibility, but wasn't using it in several places.  This results in problems when using bcolz with Python 2 for various tasks (e.g., selecting multiple columns whose names are given as unicode strings, see also #272 ).

I wrote a couple tests to test unicode handling, and everything is passing on Py2, but I'm not set up right now to test bcolz on Python 3, so hopefully someone else (like maybe the helpful Mr. Travis) can check that.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -23,6 +23,9 @@ Changes from 0.11.4 to 0.12.0
   assignments.  Fixes a VisibleDeprecationWarning with NumPy 1.10
   (@FrancescAlted).
 
+- More consistent string-type checking to allow use of unicode strings
+  in Python 2 for queries, column selection, etc. (#274 @BrenBarn)
+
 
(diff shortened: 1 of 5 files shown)


BrenBarn:
The test failure appears to be due to a change in an earlier commit that's not actually part of this PR.  Not sure what is going on there.

BrenBarn:
Okay, fixed that.  There is now a build error that's totally beyond my ken.

FrancescAlted:
Well, travis passed without problems for Python 3.4.  And appveyor is passing too, except for the 64 bit platform, which should be an unrelated issue IMO.

Before merging, it would be nice if you can update the RELEASE_NOTES.rst.  Thanks!

FrancescAlted:
Very good.

Issue #275, opened 2015-11-05
Cannot be installed with Numpy 1.10 due to incorrect package comparison operator


grahamc:
Hi, installing `bcolz` with Anaconda 2.4.0 fails. Anaconda 2.4.0 ships Numpy 1.10.1. Numerically, 1.10 is less than 1.7, but version-wise it is greater. Unfortunately `bcolz` uses an invalid comparison operator: 

https://github.com/Blosc/bcolz/blob/2851981f220b8a4dbc8c63ddede8e1f6dac0380c/pavement.py#L47
In short:

```
>>> '1.9.2' > '1.7'
True
>>> '1.10' > '1.7'
False
>>> '1.10.1' > '1.7'
False
```

grahamc:
This question / answer might be useful: https://stackoverflow.com/questions/1714027/version-number-comparison

```
>>> from distutils.version import StrictVersion
>>> StrictVersion('10.4.10') > StrictVersion('10.4.9')
True
```

FrancescAlted:
Finally I have decided to deprecate Paver, as everything should be doable with regular setuptools.

Issue #118, opened 2015-01-09
Version checking fails with weird version numbers


esc:
For example if numpy has a version like: ``1.10.0.dev-aaf5429``

sisp:
It could be solved using `distutils.version.LooseVersion` as done in e.g. https://github.com/PyTables/PyTables/blob/develop/setup.py.

esc:
Please feel free to create a pull-request.

PR #276, opened 2015-11-05
Compare package installed versions with desirable version via distutils.version.LooseVersion

grahamc:
- https://github.com/Blosc/bcolz/issues/118
- https://github.com/Blosc/bcolz/issues/275


FrancescAlted:
Excellent.  Thank you!

PR #281, opened 2015-11-23
BLD: Defer numpy/Cython operations in setup.py.

ssanderson:
Use a lazy cmdclass object to defer accesses to `numpy` and `Cython`
until after those modules have been installed by `setuptools` via
`setup_requires`.

Without this, the fact that we're supplying numpy and Cython in
`setup_requires` is meaningless, because we access numpy at module
scope, well before we're able to tell `setuptools` that we require
numpy.

I built this because I'm trying to remove several hacks and workarounds in the [zipline setup.py](https://github.com/quantopian/zipline/pull/867) to allow users to install zipline without needing to first install numpy.  The linked PR achieves that goal, but we fail installing `bcolz` because it tries to import numpy at module scope.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -6,6 +6,10 @@ Changes from 0.12.0 to 0.12.1
 =============================
 
 #XXX version-specific blurb XXX#
+- ``setup.py`` now defers operations requiring ``numpy`` and ``Cython``
+  until after those modules have been installed by ``setuptools``.  This
+  means that users no longer need to pre-install ``numpy`` and
+  ``Cython`` to install ``bcolz``.
 
 
 Changes from 0.11.4 to 0.12.0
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Pretty neat.  Could you please add an entry in the RELEASE_NOTES.rst file?  Thanks!

ssanderson:
@FrancescAlted sure.  Should I not be concerned with the Appveyor failure?  (It appears to be on master as well FWIW.)

ssanderson:
Updated.

FrancescAlted:
Great.  Will merge.  Regarding the Appveyor issue, it is just the Python 3.4 / 64-bit combination, so nothing related with this PR.  I don't have much time for looking into this, but you know what to do if you like the challenges ;)

ssanderson:
@FrancescAlted since the goal of this for me is to allow Zipline to install bcolz from PyPI without hacks,  any chance this could be released as 0.12.1?

FrancescAlted:
0.12.1 is out.

ssanderson:
@FrancescAlted awesome!  Thanks.

PR #282, opened 2015-12-09
Upgrade c-blosc to 1.7.0

alimanfoo:
This is a minimal first step, only includes manual upgrade to c-blosc 1.7.0.


alimanfoo:
First problem I am hitting is building bcolz on my laptop, which looks like it doesn't have AVX2 support. When I run:

```python setup.py build_ext --inplace```

...I get this error:

```
c-blosc/blosc/shuffle-avx2.c:14:4: error: #error AVX2 is not supported by the target architecture/platform and/or this compiler.
   #error AVX2 is not supported by the target architecture/platform and/or this compiler.
    ^
```

I can compile c-blosc itself fine via CMake, so I am guessing there are checks in the CMake build process that are not being made by the current bcolz setup.py.

Any suggestions for how to resolve this?

**Comment by FrancescAlted** [comment]
Well, I start to remember what's was happening when I gave up upgrading to newest C-Blosc a while ago.  Hmm, I had discussions with people at Intel on how to discover CPU features straight from Python, but we did not go too far.  Maybe your best bet is to try to use a standalone c-blosc library by using the [--blosc](https://github.com/Blosc/bcolz/blob/master/setup.py#L90) option for setup.py.

**Comment by alimanfoo** [comment]
I just found the [setup.py script for python-blosc](https://github.com/Blosc/python-blosc/blob/master/setup.py), looks like there is a workaround there for not being able to detect AVX2 support, will copy that.

**Comment by alimanfoo** [comment]
OK, bcolz builds and the test suite runs with no failures:

```
$ python -c "import bcolz; bcolz.test(heavy=True)"
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
bcolz version:     0.12.2.dev5+dirty
bcolz git info:    b'0.12.1-6-g64e54a9'
NumPy version:     1.10.1
Blosc version:     1.7.0 ($Date:: 2015-07-05 #$)
Blosc compressors: ['blosclz', 'lz4', 'lz4hc', 'snappy', 'zlib']
Numexpr version:   2.4.6
Python version:    3.4.3+ (default, Oct 14 2015, 16:03:50) 
[GCC 5.2.1 20151010]
Platform:          linux-x86_64
Byte-ordering:     little
Detected cores:    4
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
Performing the complete test suite!
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
......s......s......s......s...........................................................................................................................................................................................ss...............................................s...........................................................................................................................................................................................................ss......................................................................................................ssss......................................................................................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 966 tests in 20.300s

OK (skipped=13)
```

**Comment by FrancescAlted** [comment]
Ok. Cool!

**Comment by alimanfoo** [comment]
I'm going to do some benchmarking, check that performance is at least comparable to previous state when at c-blosc 1.4.0, before making any changes to use _ctx functions.

**Comment by alimanfoo** [comment]
Is there an official way to run existing benchmarks? I see there are scripts in the bench directory, also there is bench_asv.

**Comment by alimanfoo** [comment]
I've run the asv benchmarks, comparing this branch (blosc 1.7.0) with upstream master (blosc 1.4.1) and there's no noticeable difference. 

I've also run a [benchmark with some real genotype data](https://gist.github.com/alimanfoo/88a937ef0953f4a3c5d1) to check compression and decompression speeds, blosc 1.7.0 consistently comes out a shade slower, but the compression ratio is also slightly different, which may be related. However the difference is small so seems reasonable to push ahead with blosc upgrade.

**Comment by alimanfoo** [comment]
cc @mrocklin

I did a bit more benchmarking with dask as well, to (a) check that dask works with bcolz + c-blosc 1.7.0 and no further changes, which it should as c-blosc 1.7.0 still has the same non-contextual global locking functions as before, and bcolz is holding the GIL, and (b) to establish a performance baseline which we might hope to improve by switching bcolz to use the contextual non-locking blosc functions and also releasing the GIL. Results here: http://nbviewer.ipython.org/gist/alimanfoo/88a937ef0953f4a3c5d1/bench_bcolz_blosc_dask.ipynb

I compared three implementations of sum() over a large-ish array: bcolz's carray.sum(), the non-parallel chunked implementation of asum() from the [scikit-allel chunked module](http://scikit-allel.readthedocs.org/en/latest/chunked.html#allel.chunked.core.asum), and dask.array.sum().

I also tried this with two different arrays of data: the first an arange, the second some real genotype data.

Good news is that dask seems to work fine with bcolz + c-blosc 1.7.0 and no further changes. Also, performance is similar between c-blosc 1.4.1 and c-blosc 1.7.0.

The one slightly odd thing is that dask seems pathologically slow when computing a sum over the arange data, although it competes much better when working on real genotype data.

E.g., limiting blosc to 1 thread, for the arange data carray.sum() takes 1.37s, allel.chunked.asum() takes 2.24s, and da.sum() takes at least 4.17s even when using 4 workers.

However, for the genotype array, carray.sum() takes 1.93s, allel.chunked.asum() takes 2.58s, and da.sum() takes 2.12s with 4 workers.

I can't think why the arange data would hurt dask performance specifically, but not the genotype data.

**Comment by mrocklin** [comment]
So it's clearly not anything related to the actual summing of numbers, which is quite fast if we run from memory.

```python
In [1]: import numpy as np

In [2]: x = np.arange(100000000)

In [3]: y = np.random.random(100000000)

In [4]: import dask.array as da

In [5]: xx = da.from_array(x, chunks=1e6)
    
In [6]: yy = da.from_array(y, chunks=1e6)

In [7]: %time xx.sum().compute()
CPU times: user 270 ms, sys: 4.51 ms, total: 274 ms
Wall time: 72.2 ms
Out[7]: 4999999950000000

In [8]: %time yy.sum().compute()
CPU times: user 292 ms, sys: 2.97 ms, total: 295 ms
Wall time: 71.2 ms
Out[8]: 49996652.721008003
```

I wonder if you could test this entirely without dask, and just pull blocks from the array with a `ThreadPool`.  Perhaps Blosc is still somewhat unhappy being used from multiple threads.

mrocklin:
Also, fantastic work.  Thank you for taking this on.

alimanfoo:
I've added support for using the contextual blosc decompression function. GIL is released before calls to blosc_decompress_ctx().

I have made this optional for now, because I wanted to be able to compare both contextual and non-contextual easily, and because it might be desirable to give the user the option, rather than fixing on contextual calls with only a single thread.

Good news, I reran the [dask benchmarks](http://nbviewer.ipython.org/gist/alimanfoo/88a937ef0953f4a3c5d1/bench_bcolz_blosc_dask.ipynb?flush_cache=True) and the changes seem to have the desired effect. With the arange data, dask with 4 threads now beats my non-parallel chunked implementation of sum. With the genotype data, dask with 4 threads beats both other implementations.

mrocklin:
Hooray!

alimanfoo:
:-) Going to leave this for today. 

@mrocklin I assume that when dask is obtaining a chunk of data from a wrapped array, it uses __getitem__() with a slice? I am wondering if there is some overhead in calling __getitem__ on a carray, which is somehow magnified when the data are an arange. The carray.sum() method bypasses __getitem__ on the carray and calls __getitem__ directly on each chunk.

mrocklin:
Perhaps, there is certainly going to be a bit of Python overhead.  I don't think that this would explain the seconds of difference though, nor why it was different for arange vs real data.  

Blosc is known to behave differently on sequential data like arange.  Not all routines respond well to running well in parallel with themselves.  

In either event, from your message above it sounds like you've resolved this issue?

I'm curious, how valuable are these performance increases to you and your community for science reasons?  Have you tried these operations on larger machines (is there any motivation for this?) and if so how do they compare?

FrancescAlted:
Nice work @alimanfoo .  Also, you may want to use the lz4 compressor for your new benchmarks.  The version vendored in newer C-Blosc is quite faster than in previous versions.

FrancescAlted:
Also, another thing that you might want to try is to use several threads with blosc and several with dask for the contextual decompression.  Or this does not work?  It should...

alimanfoo:
So it looks like ``carray.__getitem__`` has some overhead which can almost double the time to run a chunk-wise computation. [This notebook](http://nbviewer.ipython.org/gist/alimanfoo/88a937ef0953f4a3c5d1/bcolz_profile_getitem.ipynb) has some profiling data for computation of a sum over a carray arange. Again I compared carray.sum() which accesses each chunk directly, my own simple chunked_sum() which has to go through ``carray.__getitem__`` to retrieve each chunk, and da.sum() which also has to go via ``carray.__getitem__``.

With current bcolz, both my chunked_sum and dask spend as much time in ``carray.__getitem__`` (``carray_ext.pyx:1855(__getitem__)``) as they do within ``chunk._getitem`` (``carray_ext.pyx:467(_getitem)``) where decompression actually occurs. 

I believe this overhead arises for two reasons. Firstly, ``carray.__getitem__`` sets up a numpy array for the result, then calls ``chunk._getitem`` which sets up it's own numpy array for the decompressed chunk data. So memory for the decompressed data is allocated and filled twice. Secondly, ``carray.__getitem__`` currently iterates over every single chunk in order to figure out which chunks actually have data needed to fulfill the request. This iteration and the calls to ``clip_chunk`` within each loop costs a bit of time. 

In the case where a simple slice is requested with no step, it is possible to store the decompressed data from each chunk directly into the outer numpy array created to hold the overall result. This saves double memory allocation. 

For slices it is also possible to figure out the first and last chunk required, and so avoid iterating over every single chunk.

After making these modifications, my chunked_sum is exactly the same speed as carray.sum(). All tests also pass. da.sum() with the synchronous scheduler is also exactly the same speed. 

(Aside: da.sum() with the threaded scheduler and only 1 worker is very slow in both cases, but that is probably a different issue.)

I have included the changes to ``carray.__getitem__`` (ce38653) in this branch to help understand the maximum potential of dask with bcolz, however if you would like to see these changes broken into a separate PR then just say.

alimanfoo:
@FrancescAlted I tried using several blosc threads with the contextual blosc calls but the results are terrible, i.e., at least an order of magnitude slower (I didn't wait any longer). I'm guessing this is because each call to blosc_decompress_ctx() will re-initialise a thread pool if numinternalthreads is > 1, and this is happening for every single chunk in the array?

alimanfoo:
@mrocklin my main scientific motivation is the desire to push the limits of interactive genetic data analysis. The genetic data I typically work with is very rich and needs to be explored from many different angles to build understanding of data quality issues and to discover interesting signals in the data.

Some analyses are compute heavy and will always interrupt the flow, however a lot of work also goes into preparation, exploration and visualisation of data which is less compute heavy, and where I/O can be as costly as compute. In these cases, if lots of steps in the workflow each take minutes rather than seconds to compute, it very much interrupts the flow and limits the number of avenues that can be explored.

The current scale of the data is ~100 million features by ~1000 observations which is not actually that big, with compressed arrays and chunked algorithms it is perfectly possible to do interactive analysis on a commodity laptop or desktop. However, the costs of genome sequencing are coming down very quickly, and we expect to be working with datasets 10 times this size within the next couple of years.

The main limitation in my analysis code right now is the fact that all computations are done in series. My desktop is quite modern and has 8 cores and it pains me to sit there for minutes waiting for a computation to finish and watching only 1 of 8 CPUs being used. I think dask is a great way for me to start parallelising a lot of the work I do. I particularly like the threaded scheduler, because it doesn't incur any of the overhead or complexity of working with multiple processes. I.e., it is very simple to set up and use, and fits very nicely with interactive analysis on a single machine. 

I do have bigger machines available but that won't help if I'm not able to effectively utilise multiple cores. Also I'm keen to push what is possible on commodity machines because it makes this type of data analysis so much more accessible to more people. With a laptop and a 2T hard disk I can sit on a train and get all kinds of things done. I also have access to a compute cluster with a job scheduler (SGE) so I can do big things in parallel if needed, but there is a lot of overhead that comes with coding things to work against a job scheduler and also monitoring and debugging results, so another reason to push what's possible on a single machine with multiple cores.

Sorry that's a bit vague, but it really has been transformative for me and my colleagues to be able to work interactively with the data we currently have, which is mainly down to efficient data structures (numpy, HDF5 and bcolz). If we can start parallelising some work via dask I think that will further accelerate current work and put us in a great place for the data that's coming next.

FrancescAlted:
Nice profiling work!  And yes, I would prefer to decouple the PR for integrating a new version of Blosc from the one that optimizes slices (separation of concerns).  Your optimization is general, so I will be happy to accept it as another PR.

alimanfoo:
Linking across to https://github.com/Blosc/python-blosc/issues/95 re AVX2 detection.

FrancescAlted:
:+1:  Very useful info.

2015-12-10 13:21 GMT+01:00 Alistair Miles <notifications@github.com>:

> @mrocklin <https://github.com/mrocklin> my main scientific motivation is
> the desire to push the limits of interactive genetic data analysis. The
> genetic data I typically work with is very rich and needs to be explored
> from many different angles to build understanding of data quality issues
> and to discover interesting signals in the data.
>
> Some analyses are compute heavy and will always interrupt the flow,
> however a lot of work also goes into preparation, exploration and
> visualisation of data which is less compute heavy, and where I/O can be as
> costly as compute. In these cases, if lots of steps in the workflow each
> take minutes rather than seconds to compute, it very much interrupts the
> flow and limits the number of avenues that can be explored.
>
> The current scale of the data is ~100 million features by ~1000
> observations which is not actually that big, with compressed arrays and
> chunked algorithms it is perfectly possible to do interactive analysis on a
> commodity laptop or desktop. However, the costs of genome sequencing are
> coming down very quickly, and we expect to be working with datasets 10
> times this size within the next couple of years.
>
> The main limitation in my analysis code right now is the fact that all
> computations are done in series. My desktop is quite modern and has 8 cores
> and it pains me to sit there for minutes waiting for a computation to
> finish and watching only 1 of 8 CPUs being used. I think dask is a great
> way for me to start parallelising a lot of the work I do. I particularly
> like the threaded scheduler, because it doesn't incur any of the overhead
> or complexity of working with multiple processes. I.e., it is very simple
> to set up and use, and fits very nicely with interactive analysis on a
> single machine.
>
> I do have bigger machines available but that won't help if I'm not able to
> effectively utilise multiple cores. Also I'm keen to push what is possible
> on commodity machines because it makes this type of data analysis so much
> more accessible to more people. With a laptop and a 2T hard disk I can sit
> on a train and get all kinds of things done. I also have access to a
> compute cluster with a job scheduler (SGE) so I can do big things in
> parallel if needed, but there is a lot of overhead that comes with coding
> things to work against a job scheduler and also monitoring and debugging
> results, so another reason to push what's possible on a single machine with
> multiple cores.
>
> Sorry that's a bit vague, but it really has been transformative for me and
> my colleagues to be able to work interactively with the data we currently
> have, which is mainly down to efficient data structures (numpy, HDF5 and
> bcolz). If we can start parallelising some work via dask I think that will
> further accelerate current work and put us in a great place for the data
> that's coming next.
>
> —
> <https://github.com/Blosc/bcolz/pull/282#issuecomment-163594454>.
>



-- 
Francesc Alted

FrancescAlted:
Regarding the use of more that 1 thread in the _ctx mode, I think [this PR](https://github.com/Blosc/c-blosc/pull/153) can help.  Would you mind to try it out and see?

alimanfoo:
``carray.__getitem__`` optimisation broken out to https://github.com/Blosc/bcolz/pull/283

FrancescAlted:
Hmm, this is becoming a good reference for others wanting to see what kind of performance they can expect from bcolz.  When/if you re-rerun the benchmarks at the end of the improvements, could you add a case with clevel=0?

alimanfoo:
@FrancescAlted I'm out of time for this week but could do more on this next week. Will try out the c-blosc PR regarding nthreads. Also if you think this is heading in a promising direction I could look at contextualising the compression calls as well, although I haven't given any thought to that so far.

FrancescAlted:
You mean being able to set the contextual call as the default (in case that https://github.com/Blosc/c-blosc/pull/153 works well) and deprecate the non-contextual one?  If so, and you don't detect slowdowns, I am +1.

alimanfoo:
Actually I just meant switching to use blosc_compress_ctx instead of blosc_compress, so far I have only looked at the decompression calls. In particular I wasn't sure if it would be OK to release the GIL for calls to blosc_compress_ctx? In a situation where something running multiple threads like dask is trying to store results in a carray, do we need to guard against the possibility that two threads attempt to simultaneously store data in the same carray chunk? Or is this not a worry, because it is only for the blosc_compress_ctx call that we release the GIL, so everything else is effectively synchronised by the GIL? 

@mrocklin can you see any potential issues with da.store() if bcolz releases the GIL for calls to blosc_compress_ctx?

FrancescAlted:
Ah, I see.  Well, every chunk is split internally in blocks so that Blosc can independently compress each of them, so I don't see any problem in having concurrent threads compressing different chunks.  I would give it a try, and in case this breaks, one can always fall-back to the current approach.

alimanfoo:
OK, cool, I'll see what happens.

I've been trying to think of a situation in which two threads might attempt
to store data into the same chunk of a carray. I wondered if that could
happen with dask because how dask chunks a computation may not match with
how data are chunked in the destination carray. So although although dask
always stores data into non-overlapping slices of the destination array,
there may be chunks of the array that span the slice boundaries. Not sure
if any of this is making sense, will defer to @mrocklin who will have
better knowledge of how the dask store() function works.

On Thu, Dec 10, 2015 at 3:37 PM, FrancescAlted <notifications@github.com>
wrote:

> Ah, I see. Well, every chunk is split internally in blocks so that Blosc
> can independently compress each of them, so I don't see any problem in
> having concurrent threads compressing different chunks. I would give it a
> try, and in case this breaks, one can always fall-back to the current
> approach.
>
> —
> <https://github.com/Blosc/bcolz/pull/282#issuecomment-163661544>.
>



-- 
Alistair Miles
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Tel: +44 (0)1865 287721

mrocklin:
Yup, that could definitely be an issue.  Dask.array implicitly trusts the underlying store to be able to handle concurrent non-overlapping `__setitem__` calls.  It might be reasonable for `bcolz` to hold an internal lock in its `__setitem__` call to ensure thread safety.

alimanfoo:
I am thinking that each carray should have it's own lock object (instance
of threading.Lock) which should be acquired within any method modifying
underlying data, which I think comprises __setitem__, append, trim and
resize. This means that the GIL could safely be released for calls to
blosc_compress_ctx, allowing other threads to carry on, but preventing
simultaneous attempts to modify data. I'm not hugely experienced with this
sort of thing so advice very welcome.

On Thu, Dec 10, 2015 at 5:48 PM, Matthew Rocklin <notifications@github.com>
wrote:

> Yup, that could definitely be an issue. Dask.array implicitly trusts the
> underlying store to be able to handle concurrent non-overlapping
> __setitem__ calls. It might be reasonable for bcolz to hold an internal
> lock in its __setitem__ call to ensure thread safety.
>
> —
> <https://github.com/Blosc/bcolz/pull/282#issuecomment-163700535>.
>



-- 
Alistair Miles
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Tel: +44 (0)1865 287721

mrocklin:
In principle that would help by increasing the granularity at which we could concurrently access the same data.  In practice I've never used thousands of locks.  I don't know what the performance implications of that would be.  It'd be an interesting experiment.

alimanfoo:
I don't think you'd need thousands of lock objects, just one per carray
object, to prevent concurrent writes on that carray. Or did you mean
something different?

On Thursday, 10 December 2015, Matthew Rocklin <notifications@github.com>
wrote:

> In principle that would help by increasing the granularity at which we
> could concurrently access the same data. In practice I've never used
> thousands of locks. I don't know what the performance implications of that
> would be. It'd be an interesting experiment.
>
> —
> <https://github.com/Blosc/bcolz/pull/282#issuecomment-163710483>.
>


-- 
Alistair Miles
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Tel: +44 (0)1865 287721

mrocklin:
Ah, yes, that makes more sense.  I thought that you were talking about one lock per blosc block.

FrancescAlted:
Well, the typical write access that favors bcolz is in appends, and there is a chunk buffer that when it is filled, it is compressed, so frankly, I don't see an immediate advantage in a contextual compression because the compression step should be serialized in some way or another.  So I'm -0 now on this (but an experiment won't hurt).

alimanfoo:
Hi @FrancescAlted, I don't think Blosc/c-blosc#153 is going to help at all with ``blosc_decompress_ctx()``. If I am understanding things right, that PR only changes ``blosc_set_nthreads()`` which is relevant only to managing the pool of threads used by the non-contextual functions. Calling ``blosc_decompress_ctx()`` never touches ``blosc_set_nthreads()``, but rather goes via ``blosc_set_nthreads_(struct blosc_context* context)`` instead, which launches a new thread pool for every context.

alimanfoo:
Just to confirm, applying Blosc/c-blosc#153 doesn't help, system still hands if call blosc_decompress_ctx with >1 thread.

FrancescAlted:
I see.  Well, if you spot the way to avoid the thread pool re-creation, I'll be happy to include that.

alimanfoo:
The only way I can think of would be to allow a non-global context to
"borrow" threads from the global pool, and only create it's own if there
are none available in the global pool. Unfortunately my C skills are not up
to much, and I'm not sure how that would even work in a multithreaded
environment with multiple contextual calls being made concurrently.

FWIW I think it's a reasonable option to make the bcolz default to use the
global blosc calls, so nothing will change for existing bcolz users working
in non-threaded environments. Then we add a switch
(bcolz.blosc_use_context()) so multi-threaded programs can put bcolz into
contextual mode. When in contextual mode, bcolz only ever requests a single
blosc internal thread. This gives very good performance with dask and
avoids any hierarchies of threads.

On Fri, Dec 11, 2015 at 8:17 AM, FrancescAlted <notifications@github.com>
wrote:

> I see. Well, if you spot the way to avoid the thread pool re-creation,
> I'll be happy to include that.
>
> —
> <https://github.com/Blosc/bcolz/pull/282#issuecomment-163871998>.
>



-- 
Alistair Miles
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Tel: +44 (0)1865 287721

FrancescAlted:
I agree with your analysis.  +1 for a `bcolz.blosc_use_context()`.

alimanfoo:
Just starting to look at modifying & compressing data and noticed there is room for optimisation within ``carray.__setitem__``, have submitted this separately as #284.

alimanfoo:
Hi @FrancescAlted, if it's ok with you I'd like to split this PR into two.
The first would deal with just the c-blosc upgrade. The second would build
on the first to investigate use of contextual blosc functions and
opportunities for safely releasing the GIL. I think this would make things
more manageable.

I think the c-blosc upgrade is safe and requires no changes to bcolz other
than a minor change to setup.py to work around AVX2 detection. If we get
that merged in, that would also make it possible to work on other issues
requiring blosc 1.7 like supporting bitshuffle. I would be happy to start a
PR for bitshuffle support, at least to get the API changes in there and
some unit tests.

I think the work on releasing the GIL safely is less straightforward and
may take longer to get right. If that lives in a separate PR then it won't
hold up other work.

On Friday, 11 December 2015, FrancescAlted <notifications@github.com> wrote:

> I agree with your analysis. +1 for a bcolz.blosc_use_context().
>
> —
> <https://github.com/Blosc/bcolz/pull/282#issuecomment-163892447>.
>


-- 
Alistair Miles
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Tel: +44 (0)1865 287721

FrancescAlted:
Yes, I totally agree with your plan for creating different PR depending on the goal.  And it would be great if you can put some time for looking into the bitshuffle too.  Thanks!

alimanfoo:
I've split this PR into #285 which just deals with the blosc upgrade, and #286 where I will continue work on thread-safety.

PR #283, opened 2015-12-10
optimisations for carray.__getitem__ with slice and no step

alimanfoo:
As discussed in #282, these changes remove some overheads when slicing a carray without a step, as is common when computing some result chunk-by-chunk. All tests pass. Profiling data showing the improvement is here: http://nbviewer.ipython.org/gist/alimanfoo/88a937ef0953f4a3c5d1/bcolz_profile_getitem.ipynb

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -7,6 +7,10 @@ Changes from 0.12.1 to 0.12.2
 
 #XXX version-specific blurb XXX#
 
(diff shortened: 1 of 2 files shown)


FrancescAlted:
This looks great.  Could you please add an entry in the RELEASE_NOTES.rst file?

alimanfoo:
Added release notes entry, hope that sounds ok.

FrancescAlted:
Very good indeed.  Merging.

PR #284, opened 2015-12-11
Optimise carray.__setitem__ for slice with no step

alimanfoo:
Currently there is overhead within ``carray.__setitem__`` when replacing one or more entire chunks. Overhead arises because the loop is over all chunks, checking for overlap with clip_chunks, and because the contents of chunks with data to be replaced are always read first, even if the entire chunk's data is going to be replaced. Modifying the for loop and checking to see if an entire chunk can be replaced removes this overhead. This can cut the time for ``__setitem__`` calls in half. All tests pass. Profiling data here: http://nbviewer.ipython.org/gist/alimanfoo/88a937ef0953f4a3c5d1/bcolz_profile_setitem.ipynb

### DIFF
--- a/bcolz/carray_ext.pyx
+++ b/bcolz/carray_ext.pyx
@@ -2035,7 +2035,7 @@ cdef class carray:
         """
         cdef int chunklen
         cdef npy_intp startb, stopb
-        cdef npy_intp nchunk, keychunk, nchunks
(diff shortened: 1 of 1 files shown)


FrancescAlted:
Looks good to me.  Thanks!

Issue #131, opened 2015-02-02
fromdataframe column number issues


rnelsonchem:
Hi all. I'm trying out this module in Python3 (using Anaconda installer on Windows), and I'm having some troubles with DataFrame input. It seems if the number of columns is greater than 254, then you get an error. Example below. Change the ncols to < 255 and this works fine. Any thoughts on what I can do about this?

```
import bcolz
import string
import numpy as np

ncols = 300
nrows = 100
def namer():
    return ''.join(np.random.choice(list(string.ascii_letters), size=6))
col1 = np.array([[namer(),] for i in range(nrows)])
col2 = np.array([[namer(),] for i in range(nrows)])

names = [namer() for i in range(ncols-2)]
datacols = np.random.rand(100,ncols-2)
df = pd.DataFrame(datacols, columns=names)

df['AAA_0'] = col1
df['AAA_1'] = col2
df = df.reindex_axis(sorted(df.columns), axis=1)

bcolz.ctable.fromdataframe(df)
```
Produces the following error:
```
  File "<string>", line 12
SyntaxError: more than 255 arguments
```
I can "trick" the ctable formation (using `df` from above):
```
data = bcolz.ctable.fromdataframe(df.iloc[:,[0]])
for col in df.iloc[:,1:]:
    data.addcol(df[col].values, name=df[col].name, dtype=df[col].dtype)
```
But now the dtypes of the columns are not equivalent:
```
In : data[0]
Out: (b'MFBTAz', 'MDrfoV', 0.22661330919358336, ....
```

esc:
This only happens on python 3 it seems...

esc:
The 255 vs. 254 seems to be due to an unsigned single byte integer being used at some point, and that is too small.

jseabold:
Also running in to a similar error for a table with > 255. columns.

jseabold:
For me this has to do with use of namedtuples in ctable. You can't have a function with more than 255 arguments.

https://github.com/Blosc/bcolz/blob/master/bcolz/ctable.py#L266

FrancescAlted:
Fixed by PR #303.

PR #303, opened 2016-05-11
Don't use named tuple bc > 255 columns

jseabold:
A potential fix for #131.


jseabold:
There's a little bit of room to DRY things up, but this should keep all of the functionality of the namedtuple name validation without the 255 column name limit.

jseabold:
Appveyor failure looks to be during the build and unrelated to these changes.

FrancescAlted:
Yeah, the 255 column name limit in Python is pretty unfortunate.  Thanks for this one!

PR #304, opened 2016-05-16
Release gil for Blosc operation

FrancescAlted:


FrancescAlted:
This has been tested with the `bench/multithreaded_copy.ipynb` notebook and dask can use multiple threads safely and without no locks at all.  Merging.

CarstVaartjes:
nice! any benchmarks available?

FrancescAlted:
Yup: http://nbviewer.jupyter.org/github/Blosc/bcolz/blob/master/bench/multithreaded_copy.ipynb

It is a simple one, but it is clear that the GIL is getting released for a multi-threaded application.

CarstVaartjes:
Nice!!

PR #311, opened 2016-06-29
BUG: todataframe with columns specified

thequackdaddy:
Hello,

In `ctable.todataframe`, when you specify `columns` and `orient='columns'`, pandas checks that the length of the generator is less than the number of columns. However, generators don't have lengths so it crashes. This just limits the generator the requested columns to get around that.


FrancescAlted:
Great.  Thanks.

Issue #344, opened 2017-04-30
ctable.todataframe() does not save dataframe index values


rytsim:
Taking a pandas dataframe with a timedate index and converting it to ctable then converting it back to a dataframe the index values are lost. 

```python
df.head()
                       bid  offer
time                             
2017-04-28 05:10:35  154.0  154.5
2017-04-28 08:10:00  154.0    0.0
2017-04-28 08:10:00  154.0  155.5
2017-04-28 08:10:00  154.0  155.5
2017-04-28 08:15:06  154.5  155.5
```

Converting to a ctable and back.
```
ct = bcolz.ctable.fromdataframe(df)
df2 = ct.todataframe()
```

New dataframe does not have the index information.
```
df2.head()

     bid  offer
0  154.0  154.5
1  154.0    0.0
2  154.0  155.5
3  154.0  155.5
4  154.5  155.5
```

I could store the index values as a column, but it would make more sense to store the index values automatically.

FrancescAlted:
Yes, the index is a special column that is not represented in ctable containers, so roundtrips are currently not possible.  Perhaps adding some special attribute to designate the index column in a ctable could be the way to support this.  The attribute could be named something like: `dataframe_index_column` (or similar). Interested in sending a PR?

PR #345, opened 2017-05-02
index_as_columns option in ctable.fromdataframe() 

rytsim:
Improves on [#344](https://github.com/Blosc/bcolz/issues/344)

Makes df -->ct-->df rountrip possible.

* Added an option `index_as_columns` to save the index of a dataframe in `ctable.fromdataframe()` method.
* Changed the `ctable.todataframe()` method to check for `dataframe_index_columns` attribute and if found set the dataframe index to this column before returning it.

### DIFF
--- a/bcolz/ctable.py
+++ b/bcolz/ctable.py
@@ -644,13 +644,17 @@ class ctable(object):
         return ccopy
 
     @staticmethod
-    def fromdataframe(df, **kwargs):
+    def fromdataframe(df, index_as_columns=False, **kwargs):
         """Return a ctable object out of a pandas dataframe.
 
         Parameters
         ----------
         df : DataFrame
             A pandas dataframe.
+        index_as_columns : boolean
+            If True, the index values will be converted to columns and the
+            names of the columns will be stored in the attribute 
+            'dataframe_index_columns'.
         kwargs : list of parameters or dictionary
             Any parameter supported by the ctable constructor.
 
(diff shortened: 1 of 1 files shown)


bcolz/ctable.py
```diff
@@ -644,13 +644,17 @@ def copy(self, **kwargs):
         return ccopy
 
     @staticmethod
-    def fromdataframe(df, **kwargs):
+    def fromdataframe(df, index_as_columns=False, **kwargs):
```
FrancescAlted: Although backward incompatible, I'd say that setting `index_as_columns` to `True` as default makes more sense.

bcolz/ctable.py
```diff
@@ -706,9 +710,38 @@ def fromdataframe(df, **kwargs):
...
+                        maxitemsize = pd.lib.max_len_string_array(vals)
+                        col = bcolz.carray(vals,
+                                           dtype='U%d' % maxitemsize,
+                                           **ckwargs)
+                    elif inferred_type == 'string':
+                        maxitemsize = pd.lib.max_len_string_array(vals)
+                        # In Python 3 strings should be represented as Unicode
+                        dtype = "U" if sys.version_info >= (3, 0) else "S"
+                        col = bcolz.carray(vals, dtype='%s%d' %
+                                           (dtype, maxitemsize), **ckwargs)
```
FrancescAlted: I think that 'unicode' and 'string' branches can be fused a bit more.

FrancescAlted:
Looks good to me; I have added a couple of comments though.  Also, a couple of test units would be needed.

Thanks!

bcolz/ctable.py
rytsim: I had the backward compatibility in mind, but I do agree that `index_as_columns` to `True` as a default makes more sense.

Issue #319, opened 2016-09-15
use try/except when reading chunk instead of using os.path.exists


bordingj:
I found that a significant amount of time is spend in built-in method nt.stat() when reading in a carray. The nt.stat() is used by os.path.exists() which is called every time a chunk is read. I suggest to use try/except instead of instead of checking whether the chunkfile exists everytime a chunk is read.

FrancescAlted:
That's a nice suggestion.  Could you send the profile?  As always, a PR is very welcome.

Issue #318, opened 2016-09-13
use faster json library if available


bordingj:
attrs are currently read and written using the standard json library, which is very slow compared to libraries such as ujson (https://github.com/esnme/ultrajson) and pyrapidjson (https://github.com/hhatto/pyrapidjson). I suggest that bcolz should use those if they are available

FrancescAlted:
Nice suggestion.

PR #346, opened 2017-05-07
EAFP when reading schunk and use ujson if avaible

bordingj:
#318
#319

### DIFF
--- a/bcolz/attrs.py
+++ b/bcolz/attrs.py
@@ -11,7 +11,10 @@ from __future__ import absolute_import
 
 import os
 import os.path
-import json
+try:
+    import ujson as json
+except:
+    import json
 from .py2help import dict_iteritems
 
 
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Looks good to me.  Any performance hints on how much this PR can accelerate things?  Could you add some estimates to the `RELEASE_NOTES` files?

bordingj:
I just realized that I haven't tried compiling the .pyx yet . 
The .pyx doesnt recompile because the .pyd is already in there. 
If I delete the .pyd and try to recompile it fails. I get all sorts of error.

bordingj:
In my benchmarks read-speed is ~2% faster

FrancescAlted:
Ok, so 2% is not really significant, although your modifications are not too intrusive, so I am still open to accept this PR.  Thanks!

PR #375, opened 2018-03-25
Make cpuinfo work on Tegra Jetson TX1

felixendres:
otherwise installation aborts with the exception "py-cpuinfo currently only works on X86 and some ARM CPUs."


FrancescAlted:
LGTM.  Thanks!

PR #377, opened 2018-03-29
Handling of np.datetime64 and other scalar objects that have __getitem__ attribute in _eval_blocks()

apalepu23:
The __eval_blocks() function in chunked_eval.py mishandles np.datetime64 objects since they for some reason have the __getitem__ attribute. This fix should allow users to use ctable.fetchwhere() functions off of datetime queries.

### DIFF
--- a/bcolz/chunked_eval.py
+++ b/bcolz/chunked_eval.py
@@ -243,7 +243,9 @@ def _eval_blocks(expression, vars, vlen, typesize, vm, out_flavor, blen,
                 else:
                     vars_[name] = var[i:i+blen]
             else:
-                if hasattr(var, "__getitem__"):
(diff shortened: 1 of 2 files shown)


FrancescAlted:
Your patch looks good to me.  Could you please add a test unit for the new `np.datetime64` support?

apalepu23:
Added a test (test06) as part of the fetchwhereTest class and confirmed that it works as expected. Thanks!

apalepu23:
Removed some unused code in the test, now it should be all set - thanks!

apalepu23:
Thanks Francesc - any idea when this might be released to be pip installable for others on the team needing this? No rush, just wondering thanks!

PR #396, opened 2019-06-24
Fix compatibility issue of xrange func in ctable tutorial notebook.

bingyao:
Changes made in this PR:

`xrange` is no longer existed in Python 3, to make PY2 & PY3 compatible for running `docs/tutorial_ctable.ipynb`, I made an alias for `xrange` function in PY2:

```python
import sys
if sys.version_info[0] == 2:
    range = xrange
```


FrancescAlted:
LGTM.  Thanks!

